PR #3, opened 2017-11-02
return image from visualize for local storage / web rendering

pannous:


waleedka:
* Returning the image causes Jupyter Notebook to print an additional line before every graph because Jupyter tries to print the return value of function calls. To get around this, I'd suggest adding an optional argument, `return_image` that default to False, but if it's True then the function returns the image. 

* I'm not sure why print() as a method of an class causes a conflict. It could be a Python 2.7 issue. Are you running this on 2.7? I'd imagine it might need a lot more changes than just this one.

pannous:
- I could split display_images into another function "generate_masked_image" without plotlib side effects.
- 'print()' I was a bit puzzled myself. 'show()' would be a better name b.t.w.

waleedka:
> I could split display_images into another function "generate_masked_image" without plotlib side effects.

I think you meant display_instances(), right? I think splitting the function is a good idea. One function generates the image, and the other draws it on the matplotlib axis.

> 'print()' I was a bit puzzled myself. 'show()' would be a better name b.t.w.
Yes, I think show() or display() are better. If you make the change, please make it on all the files that call this function (including the notebooks) and I'll merge it. 

Thanks in advance.

PR #20, opened 2017-11-03
config.print not compatible with python2 -> rename to display

pannous:


waleedka:
The commit missed one of the files that still called the old print() function. I went ahead and fixed the issue on my side, it was quicker than asking you to update the missed file. Thanks for brining the issue up.

PR #45, opened 2017-11-11
Update coco.py

llltttppp:
When resetting class_ids to decide which classes should be involved in dataset, the annotations should also only load these classed.


waleedka:
Merged. Thank you!

PR #57, opened 2017-11-13
fix cv2 function calling problem

np-csu:
function cv2.rectangle, cv2.circle and cv2.fillPoly returns None.


waleedka:
Hmm. If these functions returned None then the code wouldn't work. But it does work in my tests, so I'm guessing there is another issue. Which version of OpenCV are you using? I'm using 3.1

np-csu:
@waleedka I'm using OpenCV 2.4.13.

here is the function description from cv2.rectangle :
Docstring: rectangle(img, pt1, pt2, color[, thickness[, lineType[, shift]]]) -> None

And I find a server with OpenCV3, the description of cv2.rectangle is different:
Docstring: rectangle(img, pt1, pt2, color[, thickness[, lineType[, shift]]]) -> img

waleedka:
I see, thanks! Your change makes it work for both versions of OpenCV. Merged.

PR #68, opened 2017-11-17
fix inspect_data error by adding gt_class_ids

PaulChongPeng:
There is a bug in inspect_data.ipynb.
The length of list which is used to receive inputs from data_generator is incorrect.
I add gt_class_ids in the inputs list to fix it.


waleedka:
Thank you!

PR #75, opened 2017-11-20
Fix word & Feat: If multiple anchors have the same IoU.

JNingWei:


JNingWei:
Fix word

JNingWei:
Feat: If multiple anchors have the same IoU.

waleedka:
I think the update to match multiple anchors is not correct. Have you tested it?

JNingWei:
@waleedka 
Sorry, I just wrote a simple array to test it. Maybe my test method is not entirely correct.

JNingWei:
@waleedka 
I re-tested my code carefully and found that I was wrong. Thank you for your reminder, I will be carefully tested next time.

waleedka:
Thank you! I cherry-picked the one commit that has the spelling error and merged it.

PR #101, opened 2017-11-27
Fix variable typo & Fix bug: cannot display instances

JNingWei:


waleedka:
Thank you for the variable fix. I can merge that. 

But the other change in display_instances(), it's not clear to me why it's needed. The function works and displays instances already. Are you encountering a special case where adding plt.show() is needed?

JNingWei:
@waleedka
Yeah, only by adding *plt.show()* can normally display the instances.
I use:
Python 3.6.3
matplotlib .__ version__ == 2.1.0

JNingWei:
@waleedka 
Maybe bug for different version? Could you please tell me why?

jmtatsch:
@JNingWei I guess waleedka uses display_instances() only from the jupyter notebook. Because he uses %matplotlib inline there, an additional plt.show() is not needed. Do the notebooks still run properly with plt.show()?

waleedka:
@jmtatsch You're right, I've only called display_instances() from Jupyter notebooks. And I assumed, incorrectly it seems, that others would use it the same way. Can you give me examples of how you use these visualization functions? Do you call them from a regular Python script? Or from a different environment similar to Jupyter Notebook?

jmtatsch:
Jupyter notebooks are great for reproducible research but after that people may want to call your ( incredibly beautiful) visualisation methods directly from their scripts. By the way, an additional plt.show() doesn't seems to hurt a jupyter notebook, i tried it out - except for maybe the canvas is redrawn once more. I think you can safely merge this PR.

waleedka:
@jmtatsch @JNingWei Thank you both, for the PR and the clarification. It's merged.

PR #107, opened 2017-11-29
fix calling the baseclass methods with python3

jmtatsch:
this didnt happen for me with the coco dataset itself because there were always class ids, but for another dataset it blew up.


waleedka:
Thanks!

PR #120, opened 2017-12-07
Automatically download trained model file

philferriere:
Not everyone seems to have the patience to read the entire readme file before using the notebooks. If they missed the part on downloading `mask_rcnn_coco.h5`, make it easy on them -- just download the file so they can get started right away. Tested on Windows and Ubuntu.


waleedka:
This is a great addition. Thank you!

A couple of comments:

1. It might be better to make the download code into a function in utils.py so when new trained weights are available we just update the URL in one place. Also to reduce code duplication.
2. Jupyter Notebooks embed the images and other outputs in the file, which makes it hard to find which changes are relevant and which are not. It would be great to check-in the relevant changes only and skip the other changes in outputs and images.

philferriere:
Good notes, @waleedka. Will address those this week.

philferriere:
Done.

Just a quick note, @waleedka, to point out that the notebooks should be regenerated. There have been significant changes made in the code base:

windiff changes


These supposedly impact the confidence probabilities showed on the image outputs in the notebooks (at the least). Those are legit changes that should appear in the checked-in notebooks.

The bigger problem might be that the notebooks don't set a random seed when loading up sample images. Hence the often large deltas created whenever notebooks are checked out, run, and checked back in.  I would highly recommend that you do set a seed but also check in an updated version of the notebooks that shows the new outputs, following the recent code changes.

waleedka:
Thanks, Phil! This is merged now. I cherry-picked the last commit only since the others were reverted. 

And, yes, I agree on regenerating the notebooks. I'll try to do it this holiday season. I'll try fixing the random seed as well, that's a great idea.

PR #126, opened 2017-12-08
Automatically download COCO image files and annotations, if requested

philferriere:
If requested, automatically download annotations and image files for the COCO dataset, including the "minival" and "valminusminival" annotations provided by Ross Girshick on Dropbox. Tested with `inspect_model.ipynb` and on the command line with `python coco.py evaluate --dataset=cocodataset --model=mask_rcnn_coco.h5 --download=True` on a machine without the dataset. The latter populates the folder `cocodataset` with the correct image data and annotation files.

### DIFF
--- a/coco.py
+++ b/coco.py
@@ -64,6 +64,7 @@ DEFAULT_DATASET_YEAR = "2014"
 #  Configurations
 ############################################################
 
+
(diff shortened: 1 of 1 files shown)


waleedka:
This is a really cool addition. Thank you.

PR #178, opened 2018-01-09
correct data type issue of scale and image size

hariag:
data type of scale should be float, and data type of image size should be integer


hariag:
@waleedka Hi, I noticed your code working on python3 correctly, this patch make it working on python2.7.

ps48:
@hariag  Rather than float division at different places we could have just used
 `from __future__ import division `

hariag:
@ps48 thanks.

PR #233, opened 2018-02-06
Refactor compute AP using mask instead of bounding boxes

DingkunLiu:
In original paper, the model is evaluated by mask AP. In this pull request, overlapping computation with masks was implemented and replaced the original steps using bounding boxes.


waleedka:
Thank you!

PR #239, opened 2018-02-07
Create dataset class for Coco Stuff

monomon:
The official Coco Stuff dataset follows the same format as used by pycocoapi.
With recent work on the coco api it is now possible to work with the Stuff dataset in Mask_RCNN.

Here I am adding a `CocoStuffConfig` and a `CocoStuffDataset` class for loading Coco Stuff.
The included training and evaluation routines have been tested, but not extensively yet.

I am interested in contributing long-term to make the `CocoDataset` class compatible with both sets and removing this additional file. Would you agree this would be beneficial?

One thing to note is how I am loading the Coco categories as well. This is not very pretty and needs a better solution. The idea is that I might decide to train on both datasets (e.g. at different times) so the learned labels have to be consistent.

### DIFF
--- a/cocostuff.py
+++ b/cocostuff.py
@@ -1,0 +1,290 @@
+"""
+Adapted for the COCO Stuff dataset from the base COCO code.
+https://github.com/nightrome/cocostuff
+
+Tested with COCO Stuff 2017 from
+http://cocodataset.org
+
+Note: This class should most likely eventually disappear, once all differences
+between the coco and stuff datasets are bridged.
+
+Note: This script relies on the nightrome fork of cocoapi:
+https://github.com/nightrome/cocoapi
+"""
+
+import os
+import time
+
+from pycocotools import coco as pycoco
+from pycocotools.cocostuffeval import COCOStuffeval
+
+# import zipfile
+# import urllib.request
+# import shutil
+
+import coco
+import model as modellib
+
+
+# Root directory of the project
+ROOT_DIR = os.getcwd()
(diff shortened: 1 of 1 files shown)


waleedka:
This is great! Thank you. I'll do a quick test and merge it. You're okay with your work being under the MIT License, right?

monomon:
Yes, sure. Feedback is really appreciated! Do you think the way Coco categories are loaded should work?

Currently am getting weird results with it. I've trained only a few thousand examples, so maybe I need to be more patient. (The Mask R-CNN paper authors used 160k, which I'm far from)

waleedka:
I haven't looked at the code yet. But, to be honest, I'm not sure how Mask RCNN would handle `stuff`.  It's designed to find objects, so it might not do too well. 

Have you read the StuffNet paper? They use a different approach to combining objects and stuff.
https://arxiv.org/abs/1610.05861

monomon:
Hm I had not seen that, thank you. Will give it a try.

It seems to me it's more or less the same problem, so I would expect it to work, though :)

monomon:
By the way, the [stuff-10k dataset](https://github.com/nightrome/cocostuff) contains *all* categories. Any idea why the challenge 45k dataset only has the stuff ones? Seems like a weird decision to me.

rgommers:
@monomon did you continue on this, and get some good/usable results in the end?

monomon:
Hey, sorry for not replying sooner. My ML project is on hold currently, but will continue with it soon enough.
The latest results I had were somewhat encouraging, but were never great.

Issue #159, opened 2017-12-27
I have nan rpn bbox loss


dhkwon1122:
Hi, @waleedka 

I'm trying to fine-tune imagenet pre-trained model to coco task(detection and segmentation) with your code.
I already tried sample train, and to get more precise model, I start with my own pre-trained model(exactly resnet-101).
But I got nan at rpn bbox loss. the other loss seems not that bad,

what can be suspect of this phenomenon?
anything is okay, please give me advice.

Thank you.

Liron-2:
I have this problem too and this error before: 

model.py:1393: RuntimeWarning: divide by zero encountered in log
  np.log(gt_h / a_h)

hanskrupakar:
I have the same problem as well as divide by zero in log.

```
  3/3000 [..............................] - ETA: 149:10:52 - loss: nan - rpn_class_loss: nan - rpn_bbox_loss: nan - mrcnn_class_loss: nan - mrcnn_bbox_loss: 0.5063 - mrcnn_mask_loss: 0.2470
   4/3000 [..............................] - ETA: 156:12:27 - loss: nan - rpn_class_loss: nan - rpn_bbox_loss: nan - mrcnn_class_loss: nan - mrcnn_bbox_loss: 0.3798 - mrcnn_mask_loss: 0.1853
   5/3000 [..............................] - ETA: 190:53:22 - loss: nan - rpn_class_loss: nan - rpn_bbox_loss: nan - mrcnn_class_loss: nan - mrcnn_bbox_loss: 0.3038 - mrcnn_mask_loss: 0.1482
```

heidi-z400:
I was getting a loss of nan. I had 'NUM_CLASSES' set to the number of classes in my training data. 
I added +1 to that for 'BG' (background) and it fixed the problem.

viig99:
I am also getting the same issue, with another dataset
      "Configurations:\n",
      "BACKBONE_SHAPES                [[64 64]\n",
      " [32 32]\n",
      " [16 16]\n",
      " [ 8  8]\n",
      " [ 4  4]]\n",
      "BACKBONE_STRIDES               [4, 8, 16, 32, 64]\n",
      "BATCH_SIZE                     8\n",
      "BBOX_STD_DEV                   [0.1 0.1 0.2 0.2]\n",
      "DETECTION_MAX_INSTANCES        100\n",
      "DETECTION_MIN_CONFIDENCE       0.7\n",
[...]

PR #268, opened 2018-02-21
fix nan rpn bbox loss issue

lianyi:
hopefully fix some of the cases in #159

The cropping will cause boxes to be zero thus the division zero issue as reported in #159. This fix at lease fix the same issue I had.

### DIFF
--- a/model.py
+++ b/model.py
@@ -1205,7 +1205,7 @@ def load_image_gt(dataset, config, image_id, augment=False,
 
     # Note that some boxes might be all zeros if the corresponding mask got cropped out.
     # and here is to filter them out
-    _idx = np.sum(np.where(mask > 0, 1, 0), axis=(0, 1)) >= 0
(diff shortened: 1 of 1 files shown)


waleedka:
Thanks! I have a couple of questions about this before I merge it in:

```
_idx = np.sum(np.where(mask > 0, 1, 0), axis=(0, 1)) >= 0
```

* The last` >=` should be `>`, right? Otherwise, masks with zero pixels are included.
* The mask at this stage is still a binary mask, so `np.where(mask > 0)` is probably not needed. 

Please let me know if I missed something.

lianyi:
Thanks, @waleedka ! 
Yes, the last `>=` should be `>` to make sure the mask is not empty.   And yes `np.where(mask > 0)`  , if it's still a binary mask. PR updated.

waleedka:
Thank you! Merged.

Issue #102, opened 2017-11-28
StopIteration


123liluky:
When I train net on 35kval images, StopIteration happens in the 295th step  of 6th epcho, like below.
295/1000 [=======>......................] - ETA: 11:56 - loss: 1.3895 - rpn_class_loss: 0.0333 - 
rpn_bbox_loss: 0.5186 - mrcnn_class_loss: 0.3381 - mrcnn_bbox_loss: 0.2326 - mrcnn_mask_loss: 0.2669
Traceback (most recent call last):
File "coco.py", line 397, in <module>  layers='heads')
File "D:\work\code\matterportMask_RCNN\Mask_RCNN-master\model.py", line 2183,in train
  **fit_kwargs
File "G:\Anaconda3\lib\site-packages\keras\legacy\interfaces.py", line 87, in wrapper
  return func(*args, **kwargs)
File "G:\Anaconda3\lib\site-packages\keras\engine\training.py", line 2083, in fit_generator
  generator_output = next(output_generator)
StopIteration
I can't figur out what happened, can you help me? Changes I made in config.py BATCH_SIZE=1,IMAGE_MAX_DIM=512,IMAGE_MIN_DIM=400.

waleedka:
I see a `StopIteration` error being raise. That seems to indicate that the data generator is failing for some reason. Maybe you have images missing in your dataset, or any other issue loading the data. 

Use the `inspect_data` notebook and verify that the data generator loads images correctly. Specifically, there is a block of commented out code that calls next() on the data generator repeatedly a lot of times to verify it's working. Try to run it on the entire dataset and verify that all images can be loaded successfully. 

Also, make sure you get the latest version of the code from github.

123liluky:
I have several other questions: 
1. When I try to train on 35k val images, I just remove “dataset_train.load_coco(args.dataset, "train")” in coco.py. Must I do some other changes?
2. When I set IMAGE_MAX_DIM=512,IMAGE_MIN_DIM=400 in config.py, I also set RPN_ANCHOR_SCALES = (16, 32, 64, 128, 256). Are there other parameters closely related to INPUT IMAGE SIZE?
Thank you!

waleedka:
Yes. That's the mean pixel for the COCO dataset.

jmtatsch:
@123liluky @waleedka those are the imagenet means not the coco means

123liluky:
[...]

Issue #100, opened 2017-11-27
why val_loss is getting higher and higher？


zbsean:
why val_loss is getting higher and higher？	
How to choose the optimal model

waleedka:
Since you haven't provided any details, the only thing I could suggest is lowering the learning rate and trying again.

PR #306, opened 2018-03-07
Mobile Mask RCNN

Cpruce:
Actual commit is bf533ea.

Images show inference on GTX 1080.

Seems to have a memory leak. Haven't focused on that yet.
**EDIT: Memory leak may be a bug in tensorflow https://github.com/tensorflow/tensorflow/issues/13359** 

Let me know how you'd like to deal with the model I produced.

Thanks!
Cory


ps48:
@Cpruce I went through the commit superficially, just wanted to ask if you have trained the mobilenet with pretrained-imagenet weights or not and If so can you add the fucntionality to use it using --model=imagenet flag.

Cpruce:
You can check my last commit. I haven't tested the addition to get_imagenet_weights out yet

ps48:
I just checked it, It should be mobilenet_1_0_224_tf_no_top.h5 that is to be used. "No top file" does not include the final layers like Global Avg pooling/Final Fc/Softmax. [link for mobilenet_1_0_224_tf_no_top.h5](https://github.com/fchollet/deep-learning-models/releases/download/v0.6/mobilenet_1_0_224_tf_no_top.h5) you can replace the path with this link.

Cpruce:
Yup, I've got that file. I finetuned on that to bootstrap and have been using checkpoints I've produced since. Wasn't sure which one I had got working since the one I listed was the most recent one I downloaded, which is why I mentioned I didn't test that yet :) You're probably right but we may be able to use both when we exclude layers (in this case the final dense network).

I chose the larger input sized model because I wanted to get mobilenet to be on par with the resnet50 backbone. Not surprisingly, it's proving to be non-trivial :D

John1231983:
@Cprune: your mAP for bbox is better than the last version of repo (0.37 / 0.34) for resnet 101. Could you also show the mAP for segmentation? The repo achieved 0.29. And if it is possible please share to us your pretrained of resnet 101 on coco

Cpruce:
@John1231983 I've added mobilenet224 as a backbone, as opposed to resnet101. AFAIK, you should be able to use resnet50 and resnet101 interchangeably as it seems the code supports it. Unfortunately my mAP on par with the original backbone, resnet50. See Inference Performance in README.md at https://github.com/Cpruce/Mask_RCNN. It is much faster and smaller though. The inference performance results are bbox in that image, but segm is extremely close, so the picture is representative. You can also see some coco test images on the README. However, segmentation and bounding box results were poor on the shapes images. I believe this is due to either overfitting from coco or not finetuning on shapes enough.

John1231983:
Good job @Cpruce : Could you please share your pre-train coco using resnet101? Can I use it as pretrain for another dataset for instance segmentation I am very interesting why it can provide better than matterport performance(0.34 on bbox), compared with 0.37 as you.

Cpruce:
Do you mean mobilenet224? I haven't tried resnet101 as a backbone. I'm also not sure where you're getting the numbers from. These are the results for mobilenet224 after 500 something epochs:
screen shot 2018-03-10 at 12 11 47 pm

And these are the results of resnet50:
screen shot 2018-03-10 at 12 11 30 pm
Not sure exactly how many epochs but probably much less than 500

John1231983:
Sorry. I means resnet50(second image). Have you give me your pretrain resnet 50 on coco ?

John1231983:
@Cpruce : So, you just you the matterport pretrain model and it give 0.37 AP. Am I right? Because the report of matterport shows 0.34

Cpruce:
@John1231983 I fine-tuned the model a bit on coco. Are you able to do the same? ie do you have a gpu?

John1231983:
Yes. I have 1 titanx pascal 12gb. So if I understand correctly, you start to train coco with imagenet resnet50 pretrained. You also finetune config. After training, you got a coco pretrain and the 0.37 is from it. If it is correct, I am very happy to see your coco pretrain because I would like to use it (instead of matterport's coco pretrain as you gave me)

Cpruce:
@John1231983 With the model posted above and a clean checkout of master, I consistently produce 0.37 for the mAP at the first state. Could you post what results you get from running evaluate on that model?

On a tangent, my mobilenet backbone is not as fast as I thought and a decent amount was due to the gpus. There's still a significant increase in speed, though it is still rather slow and the scores could be better. You can check the readme now for the update

John1231983:
@Cpruce: I have not enough GPU to train coco. I just use the author pretrained coco. It provided 0.29 ap in seg task. Could you provide us your pretrain coco that shown 0.37 in bbox?

Cpruce:
@John1231983 I produced the 0.37 mAP from the base model in the release, no fine-tuning. I think you should also be able to train Coco on a Titan x as I am able to on a 1080 and a Time, individually. Let's move the conversation elsewhere. Could you please email me your keras and tensorflow versions?

waleedka:
@Cpruce Thanks! This is a great addition. I'm starting to review the code. 

In keeping with the goal of keeping the repo small and light, I'd suggest removing anything that can be removed. The things that caught my eye right away are:

* Validation results can probably be included in the Releases page like we did with the previous release. Or, alternatively, I'm thinking of adding another page, other than the main README, that covers more details. 
* The sample images can either be moved to the details page if I add it (still thinking about that), or they can be included in a separate blog post or an external resource. And then, I think they're only useful if we show the same image being segmented with two different backbones next to each other so it's easy to compare.
* Training loss curves are not essential and can be removed.
* `supported_architectures.py`. Doesn't have enough code to justify being a separate file.

I'm yet to go through the code, though. I'll let you know if I have any questions.

Cpruce:
@changzhonghan Mean pixel values are artifacts of the imagenet dataset https://github.com/matterport/Mask_RCNN/issues/102

https://www.learnopencv.com/keras-tutorial-using-pre-trained-imagenet-models/

changzhonghan:
The preprocess_input function in mobilenet is differenr from resnet.The preprocess of Resnet uses the mean pixel of imagenet.But thr preprocess of mobilenet is (x/122.5-1.0)，which dont uses the mean pixel.

Cpruce:
@waleedka Thanks for the great repository and your effort! I will make your requested changes after you are done reviewing. Note that only a few commit hunks are to be committed. Somethings like the notebook I found to be broken with the addition of ARCH in config. I was planning on going through it soon to make sure everything works. Please feel free to reach out if you want to discuss anything.

@changzhonghan  I did a quick test with one of my lower checkpoints for a few epochs, but the loss kept increasing. That said, this is not a completely fair test. 

I found that the [keras mobilenet](https://github.com/keras-team/keras/blob/master/keras/applications/mobilenet.py) does use mean pixel values of imagenet, like resnet50. 
`def preprocess_input(x):
    """Preprocesses a numpy array encoding a batch of images.
    # Arguments
        x: a 4D numpy array consists of RGB values within [0, 255].
    # Returns
        Preprocessed array.
    """
    return imagenet_utils.preprocess_input(x, mode='tf')`

which leads to 

`def _preprocess_numpy_input(x, data_format, mode):
    """Preprocesses a Numpy array encoding a batch of images.
    # Arguments
        x: Input array, 3D or 4D.
        data_format: Data format of the image array.
        mode: One of "caffe", "tf" or "torch".
            - caffe: will convert the images from RGB to BGR,
                then will zero-center each color channel with
                respect to the ImageNet dataset,
                without scaling.
            - tf: will scale pixels between -1 and 1,
                sample-wise.
            - torch: will scale pixels between 0 and 1 and then
                will normalize each channel with respect to the
                ImageNet dataset.
    # Returns
        Preprocessed Numpy array.
    """
    if mode == 'tf':
        x /= 127.5
        x -= 1.
        return x

    if mode == 'torch':
        x /= 255.
        mean = [0.485, 0.456, 0.406]
        std = [0.229, 0.224, 0.225]
    else:
        if data_format == 'channels_first':
            # 'RGB'->'BGR'
            if x.ndim == 3:
                x = x[::-1, ...]
            else:
                x = x[:, ::-1, ...]
        else:
            # 'RGB'->'BGR'
            x = x[..., ::-1]
        mean = [103.939, 116.779, 123.68]
        std = None

    # Zero-center by mean pixel
    if data_format == 'channels_first':
        if x.ndim == 3:
            x[0, :, :] -= mean[0]
            x[1, :, :] -= mean[1]
            x[2, :, :] -= mean[2]
            if std is not None:
                x[0, :, :] /= std[0]
                x[1, :, :] /= std[1]
                x[2, :, :] /= std[2]
        else:
            x[:, 0, :, :] -= mean[0]
            x[:, 1, :, :] -= mean[1]
            x[:, 2, :, :] -= mean[2]
            if std is not None:
                x[:, 0, :, :] /= std[0]
                x[:, 1, :, :] /= std[1]
                x[:, 2, :, :] /= std[2]
    else:
        x[..., 0] -= mean[0]
        x[..., 1] -= mean[1]
        x[..., 2] -= mean[2]
        if std is not None:
            x[..., 0] /= std[0]
            x[..., 1] /= std[1]
            x[..., 2] /= std[2]
    return x`

Can you post what you're referencing?

changzhonghan:
In the keras.application.monilenet, the Keras-team define the preprogress_input function for mobilenet, so do inception. On the other hand, in tensorflow.slim.net, the preprogress of mobilenet and inception is different, it is (x/122.5 - 1)

changzhonghan:
This is something wrong in my previous comment. The preprocess of mobilenet is (x/127.5-1.0)，which dont uses the mean pixel of imagenet.In Keras mobilenet, they use [ imagenet_utils.preprocess_input(x, mode='tf')] which return :
if mode == 'tf':
        x /= 127.5
        x -= 1.
        return x
This function is different from resnet.

Cpruce:
You're right. I missed that return x. Perhaps the mean pixel normalization works better similar to the Coco distribution and the fixed normalization is more general. I'll test whether this performs better:

```python
def mold_image(images, config):
      """Takes RGB images with 0-255 values and subtraces
      the mean pixel and converts it to float. Expects image
      colors in RGB order.
      """
     return images.astype(np.float32)/127.5 - 1.0 #images.astype(np.float32) - config.MEAN_PIXEL  
  
  def unmold_image(normalized_images, config):
      """Takes a image normalized with mold() and returns the original."""
     return ((normalized_images + 1)*127.5).astype(np.uint8) #(normalized_images + config.MEAN_PIXEL).astype(np.uint8)
```

JonathanCMitchell:
@Cpruce  Is there an advantage to include stage 13 (block 13) of the mobilenet base in your architecture? As can be seen here [170](https://github.com/akirasosa/mobile-semantic-segmentation/blob/master/nets/MobileUNet.py#L170). The spatial dimensions of block 12 are consistent with res-nets requirements. In my experiments, I used block 12 [169](https://github.com/akirasosa/mobile-semantic-segmentation/blob/master/nets/MobileUNet.py#L169) as my `C5` output.

Cpruce:
@JonathanCMitchell I believe the advantage would be another block (more parameters) for the network to work with. Except for stage 1, I tried to remain as consistent as possible with the resnet backbone, extracting layer feature maps at the end of each pointwise convolution output shape round (64, 128, 256, 512, 1024). However, I can't say I've tried with C5 at block 12 yet.  Have you tried with C5 at block 13 and have you been able to produce results yet?

@changzhonghan So far, so good. Thank you for spotting that preprocessing discrepancy. The model with the above normalization has been converging faster
screen shot 2018-03-20 at 9 00 04 pm

JonathanCMitchell:
@Cpruce It appears that you are using trainable Batch Normalization layers in your MobileNet graph, whereas the current master branch implementation sets the BatchNorm layers to `non trainable`. You can see this where they overwrite the layer with `BatchNorm` on [56](https://github.com/matterport/Mask_RCNN/blob/master/model.py#L56)

Cpruce:
@JonathanCMitchell You're right. Thanks, I am testing `BatchNorm` now as my batch size is small. There is a noticeable improvement when training (<1.2 training loss and decreasing). I'll experiment with the Depthwise separable convolution after I return/BatchNorm change has run for a number of epochs.

ps48:
@JonathanCMitchell have you tried using Depthwise separable layers before on FPN? Can you help me understand how will it be different than TimeDistributed layer. If I understand correctly, even in the time distributed layer the convolution is done on sliced time-dimension.

JonathanCMitchell:
@ps48 The TimeDistributed layer is a means of performing the 2D convolution on each feature map from the feature pyramid network. So if the FPN contains four stages (C2, C3, C4, C5), the TimeDistributed layer is a means of looping through each stage and performing a 2D convolution on each feature map. The 2D convolution being performed is still not a DepthwiseSeparable convolution, but you can choose to make it a DepthwiseSeparable convolution. However, typically with DepthWise separable convolutions you want to include a BatchNormalization layer, but for such a small batch size it may not make sense. (My batch is is 4 images , one per GPU).

ps48:
@JonathanCMitchell Oh cool! now i get it. I initially thought you wanted to use DepthwiseSeparable layer instead of TimeDistributed layer with different depth filters to have convolutions run on (C2, C3, C4, C5). Ya we can try this approach of DepthwiseSeparable convolution instead of 2D Convolution to keep the mapping of cross-channel correlations of features intact, that we get from mobile net.

Cpruce:
screen shot 2018-03-28 at 11 23 31 pm

Unfortunately, I think that mixing the `BatchNorm` and `BatchNormalization` classes may have had a negative impact when changing the number of layers to train. I'll be a bit busy for the next few weeks but I'll try to make progress where I can and will definitely be revisiting afterwards. @ps48 @JonathanCMitchell feel free to repeat, experiment, and improve!

JonathanCMitchell:
@Cpruce Do you happen to have your training log files available? I would like to compare my results against yours. If you could email them to me that would be great jonathan@supermediafuture.com

JonathanCMitchell:
@Cpruce I noticed your backbone strides are still the same as the backbone strides of ResNet50. Although there is not 1024 channel layer. I tried changing the backbone strides to from 
`BACKBONE_STRIDES = [4, 8, 16, 32, 64]`
to
`BACKBONE_STRIDES = [2, 4, 8, 16, 32]`
To meet MobilNet's specs. However, when I do this, my rpn_bbox loss seems to increase instead of decrease.

Cpruce:
@JonathanCMitchell Thanks for continuing to experiment. I too tried [2, 4, 6, 16, 32] as the BACKBONE_STRIDES, thinking it would further decrease loss. I haven't yet looked into why the original BACKBONE_STRIDES performs better.

JonathanCMitchell:
@Cpruce Do you happen to have a run through of the inspect_weights.ipynb notebook available for this implementation? When I checked my weights, (when I set BatchNorm to non trainable inside mobilenet), I had a very large gamma value at the batch norm layer right after the first depthwise separable convolution. The value was `bn_depthwise_1/gamma:0 shape(32), Min = +0.0000 Max = 35610.7070 STD = 10044.4668.`

Cpruce:
@JonathanCMitchell I too am seeing overflow at the first depthwise convolution block batchnorm. I am also seeing some dead neurons towards the end of the model.

screen shot 2018-04-07 at 7 52 15 am
screen shot 2018-04-07 at 7 54 31 am

I'll likely experiment with different activation functions and batchnorm down the road. Maybe @waleedka or @changzhonghan has some insight on this? 

@JonathanCMitchell @GustavZ Feel free to contribute to this pull request. There's still a lot of room for improvement. If you get a mobilenetv2 backbone working as well, by all means add it to the PR or create your own. @GustavZ Have you at least done a little trial-and-error with C1-C5 assignment?

JonathanCMitchell:
@GustavZ Where are you getting the weights for MobilenetV2? Did you convert them from PyTorch? Also please let us know how your experiments turn out when you use the backbone strides associated with MobileNetV1 and V2 because they should be different than ResNet's strides. 
Specifically. `BACKBONE_STRIDES = [4, 8, 16, 32, 64]` to `BACKBONE_STRIDES = [2, 4, 8, 16, 32]`

GustavZ:
@JonathanCMitchell
1.  i dont really understand what you mean, as the spatial resolutions of mobileNetV2 are different from resnet50. they are also different from mobileNetV1, so how to assign the stages? as i understand the standard spatial resultions of MNV2  to be adjusted to fit mask r-cnn? Also what do you mean with this code snipped (what are those 'res_connect' layers, and how should they be used in mvn2 )?
```
C1 = model.get_layer('res_connect_3').output
C2 = model.get_layer('res_connect_3').output
C3 = model.get_layer('res_connect_6').output
C4 = model.get_layer('res_connect_13').output
C5 = model.get_layer('res_connect_16').output
```
2.  I don'T have any MNV2 weights yet, do you know whether there are some available somewhere?
I did not train on MNV2 backbone yet, i just started with theoretically adding it into the model.py. Which ones should be the right for MNV1 and MNV2? can they use the same?

@waleedka  you did some big refactorization over the weekend, does this affect this PR too? should we update our fork?

JonathanCMitchell:
@GustavZ To connect MobileNetV1 you should use the same spatial resolution in the conv layers that are being used for ResNet. Therefore, if you have an input that is shape `(W0, H0, 3)` and you know that in ResNet the C4 layer has a spatial resolution that is, for example, `(W0//16, H0//16, 3)`, then you should look for the MobileNet layer that is also `(W0//16, H0//16, 3)`. Spatial resolution would be the `height` and `width` of the feature maps. For MobileNet V1 this layer would be `conv_pw_relu_11`. So you would have the following in your MobileNet_graph() function:
```
C4 = model.get_layer('conv_pw_relu_11') # This represents the layer that has spatial resolution (1//16) of the input.
```
You can figure out which layer it is for MobileNet V2. I named my layers differently but I can tell you that I used the Add layers as my outputs.

GustavZ:
@JonathanCMitchell thanks for this detailed explanation! I was already aware of this, my question was more concerning that I thought mnv2 has different spatial resolutions than resent or mnv2, or in your example, different fractions of the input resolution.
So I thought it is not possible to just assign the stages to a certain mvn2 layer as they would never be the exact same resolution.

JonathanCMitchell:
@GustavZ I made C5 one block up, but you should experiment between the two. I would suggest starting one block up though. Also, you should always take the last inverted residual block per each set (defined by `n` in the original paper). So if it says `n=3`, then take the third one.

Cpruce:
@JonathanCMitchell @GustavZ I think I have an explanation for why my MobileMRCNN architecture works.

1. @JonathanCMitchell For where to place the C5 bottom-up feature map, I think I'm actually right to put it at the end of the last stage (the two last blocks may have the same output spatial resolution and thus would be in the same stage). 
"There are often many layers producing output maps of the same size and we say these layers are in the same network stage.  For our feature pyramid, we define one pyramid level for each stage.  We
choose the output of the last layer of each stage as our reference set of feature maps, which we will enrich to create our pyramid.  This choice is natural since the deepest layer
of each stage should have the strongest features."
Ref: https://arxiv.org/pdf/1612.03144.pdf - see first paragraph of Bottom-up pathway

2. As for why the lateral connections linking the bottom-up and top-down feature maps works, my belief is that Waleed's implementation of top-down inherits the dimensions from the bottom-up pathway, as long as they are the next/previous power of 2. https://stackoverflow.com/questions/43602451/cannot-set-output-dim-of-conv2d-in-keras. I checked the spatial resolutions of my feature maps against that of the original resnet50 backbone and they have a depth of one less power of 2. @waleedka do you have a good explanation?

Cpruce:
screen shot 2018-04-10 at 6 29 44 pm
screen shot 2018-04-10 at 8 22 58 am

Here are my feature map depths. `conv_pw_12_relu` and `conv_pw_13_relu` highlight #1 above

chrispolo:
Hello,i use the mobilenet as a backbone to train on the pretrained weights:mask_rcnn_coco.h5,
i met this mistake:
Loading weights /root/data/mrcnn_test/mask_rcnn_coco.h5
File "coco.py", line 484, in 
model.load_weights(model_path, by_name=True)
File "/root/data/mrcnn_test/model.py", line 2402, in load_weights
topology.load_weights_from_hdf5_group_by_name(f, layers)
File "/usr/local/lib/python3.6/dist-packages/keras/engine/topology.py", line 3201, in load_weights_from_hdf5_group_by_name
' element(s).')
ValueError: Layer #2 (named "conv1") expects 1 weight(s), but the saved weights have 2 element(s).
How to fix it?

chrispolo:
@Cpruce Hello,i use the mobilenet as a backbone to train on the pretrained weights:mask_rcnn_coco.h5,
i met this mistake:
Loading weights /root/data/mrcnn_test/mask_rcnn_coco.h5
File "coco.py", line 484, in 
model.load_weights(model_path, by_name=True)
File "/root/data/mrcnn_test/model.py", line 2402, in load_weights
topology.load_weights_from_hdf5_group_by_name(f, layers)
File "/usr/local/lib/python3.6/dist-packages/keras/engine/topology.py", line 3201, in load_weights_from_hdf5_group_by_name
' element(s).')
ValueError: Layer #2 (named "conv1") expects 1 weight(s), but the saved weights have 2 element(s).
How to fix it?
and when i trained the all layers .The loss is：
999/1000 [============================>.] - ETA: 1s - loss: 1.4577 - rpn_class_loss: 0.6384 - rpn_bbox_loss: 0.3800 - mrcnn_class_loss: 0.2417 - mrcnn_bbox_loss: 0.1073 - mrcnn_mask1000/1000 [==============================] - 1071s 1s/step - loss: 1.4570 - rpn_class_loss: 0.6383 - rpn_bbox_loss: 0.3799 - mrcnn_class_loss: 0.2414 - mrcnn_bbox_loss: 0.1072 - mrcnn_mask_loss: 0.0902 - val_loss: 0.9555 - val_rpn_class_loss: 0.6032 - val_rpn_bbox_loss: 0.3485 - val_mrcnn_class_loss: 0.0038 - val_mrcnn_bbox_loss: 0.0000e+00 - val_mrcnn_mask_loss: 0.0000e+00
Epoch 2/160
   1/1000 [..............................] - ETA: 17:06 - loss: 1.0023 - rpn_class_loss: 0.6335 - rpn_bbox_loss: 0.3686 - mrcnn_class_loss: 1.8023e-04 - mrcnn_bbox_loss: 0.0000e+00 -   2/1000 [..............................] - ETA: 17:32 - loss: 0.9545 - rpn_class_loss: 0.6060 - rpn_bbox_loss: 0.3475 - mrcnn_class_loss: 9.7235e-04 - mrcnn_bbox_loss: 0.0000e+00 -   3/1000 [..............................] - ETA: 17:30 - loss: 1.6306 - rpn_class_loss: 0.6758 - rpn_bbox_loss: 0.4205 - mrcnn_class_loss: 0.0171 - mrcnn_bbox_loss: 0.2824 - mrcnn_m   4/1000 [..............................] - ETA: 17:29 - loss: 1.4638 - rpn_class_loss: 0.6440 - rpn_bbox_loss: 0.4190 - mrcnn_class_loss: 0.0130 - mrcnn_bbox_loss: 0.2118 - mrcnn_m

Cpruce:
@chrispolo Did you produce mask_rcnn_coco.h5 from imagenet?

@waleedka Could you please check the current PR and make sure that I've addressed everything?

chrispolo:
@Cpruce i didn't produce the mask_rcnn_coco.h5 from imagenet.I trained on the pre-trained coco weight.and  after  i trained without any pre-weights. I evaluate it
train_layer2
screenshot from 2018-05-14 21-54-36
and when i trained all layers:The loss is :
1000/1000 [==============================] - 1335s 1s/step - loss: 1.1289 - rpn_class_loss: 0.5755 - rpn_bbox_loss: 0.3770 - mrcnn_class_loss: 0.0110 - mrcnn_bbox_loss: 0.0913 - mrcnn_mask_loss: 0.0742 - val_loss: 0.8496 - val_rpn_class_loss: 0.5216 - val_rpn_bbox_loss: 0.3278 - val_mrcnn_class_loss: 1.9459e-04 - val_mrcnn_class_loss: 0.0000e+00 - val_mrcnn_mask_loss: 0.0000e+00
The val_mrcnn_class_loss val_mrcnn_class_loss,val_mrcnn_mask_loss: is 0
and i consulted your code,it still showed this.
Thx a lot! Hope your reply.

chrispolo:
@Cpruce train with the imgenet
the result is:
Downloading data from https://github.com/fchollet/deep-learning-models/releases/download/v0.6/mobilenet_1_0_224_tf.h5
17227776/17225924 [==============================] - 69s 4us/step
Loading weights  /root/.keras/models/mobilenet_1_0_224_tf.h5
  File "coco.py", line 485, in <module>
    model.load_weights(model_path, by_name=False)
  File "/root/data/mrcnn_test/model.py", line 2405, in load_weights
    topology.load_weights_from_hdf5_group(f, layers)
  File "/usr/local/lib/python3.6/dist-packages/keras/engine/topology.py", line 3116, in load_weights_from_hdf5_group
    str(len(filtered_layers)) + ' layers.')
ValueError: You are trying to load a weight file containing 55 layers into a model with 79 layers.

Cpruce:
@chrispolo have you taken the master from my fork? What does `python3 samples/coco/coco.py train --dataset=/path/to/coco --model=imagenet --backbone=mobilenet224v1` produce?

chrispolo:
@Cpruce ,Yeah, i had taken the master from your fork.Thx a lot!
I will train again.I want to ask a question. when I need to train with the pre_trained imagenet firstly and not coco pre-trained weights directly?
and finish the training stage1 ~stage3?
Thx your reply.

chrispolo:
@Cpruce After i trained the head with imagenet weights:
i evaluate it 
mboile
it confused me a lot! 
Hope your reply

Cpruce:
Yes, train on imagenet from scratch using --backbone=mobilenet224v1. Let me know how it goes @chrispolo

chrispolo:
yeah, I trained the head with imagenet weights
The loss is 
mobile_epoch5
when i used the demo.ipyb to test the trained weights,
it shows nothing
li just showed this 
*** No instances to display *** 
i don't know how to solve it

chrispolo:
@Cpruce yeah, I trained the head with imagenet weights
The loss is

mobile_epoch5

when i used the demo.ipyb to test the trained weights,
it shows nothing
li just showed this
*** No instances to display ***
i don't know how to solve it

hegc:
@Cpruce Hi,  I had taken the master from your fork, used the mobilenet224v1 as backbone, trained on Tesla P40, and I was confused, why the train speed is so low?
100/100 [==============================] - 1666s 17s/step - loss: 2.4812 - rpn_class_loss: 0.0993 - rpn_bbox_loss: 1.3111 - mrcnn_class_loss: 0.0253 - mrcnn_bbox_loss: 0.4110 - mrcnn_mask_loss: 0.6345 - val_loss: 1.6659 - val_rpn_class_loss: 0.0249 - val_rpn_bbox_loss: 0.7214 - val_mrcnn_class_loss: 0.0284 - val_mrcnn_bbox_loss: 0.2167 - val_mrcnn_mask_loss: 0.6745

Cpruce:
@chrispolo Sorry for the late reply. I am revisiting the PR to wrap things up. I'll need to go through all the examples/scripts/notebooks and make sure everything is sound. I will make sure demo.ipynb is one of the first I look into.

@hegc Thanks for trying it out. Let me benchmark and compare. I'll look into the speed if I can reproduce.

Cpruce:
@hegc I am actually not seeing the same.

Training on coco2014,

Resnet50 backbone w/coco weights:
`1000/1000 [==============================] - 629s 629ms/step - loss: 1.6670 - rpn_class_loss: 0.0470 - rpn_bbox_loss: 0.4027 - mrcnn_class_loss: 0.5112 - mrcnn_bbox_loss: 0.3395 - mrcnn_mask_loss: 0.3667 - val_loss: 2.3330 - val_rpn_class_loss: 0.0438 - val_rpn_bbox_loss: 0.7585 - val_mrcnn_class_loss: 0.6890 - val_mrcnn_bbox_loss: 0.4483 - val_mrcnn_mask_loss: 0.3933`

Mobilenet224v1 backbone w/imagenet weights:
`1000/1000 [==============================] - 571s 571ms/step - loss: 2.7562 - rpn_class_loss: 0.1281 - rpn_bbox_loss: 0.6720 - mrcnn_class_loss: 0.6765 - mrcnn_bbox_loss: 0.5920 - mrcnn_mask_loss: 0.6876 - val_loss: 2.9852 - val_rpn_class_loss: 0.0699 - val_rpn_bbox_loss: 0.8338 - val_mrcnn_class_loss: 0.7368 - val_mrcnn_bbox_loss: 0.6634 - val_mrcnn_mask_loss: 0.6814`

Can you take the latest master of my fork and post the comparisons if the problem persists?

Cpruce:
@chrispolo I'm not exactly sure what your issue was. If you haven't resolved it, does the notebook run fine without the mobilenet backbone?

Here is demo.ipynb output with a mobilenet backbone:
screen shot 2019-01-15 at 8 56 04 pm

Cpruce:
Hi Waleed,

Thanks for your help! With your suggestion to use --train_bn, no dead nodes or overflows appear in inspect_weights.ipynb and everything runs fine in your recommended container.

Before wrapping things up, I have a few questions:

- Why did you choose res4w_out for resnet101 in inspect_model.ipynb?
- I see in your response to an issue that you wanted to create another README. Do you think now is a good time?
- What environments should I support? Everything works fine using your docker image since that has the target versions of the dependencies but I can confirm that tf 1.11/keras 2.2.4 runs into issues with inference.

Thanks,
Cory

mrcnn/model.py
```diff
@@ -2171,14 +2550,14 @@ def compile(self, learning_rate, momentum):
             if layer.output in self.keras_model.losses:
                 continue
             loss = (
-                tf.reduce_mean(layer.output, keepdims=True)
+                tf.reduce_mean(layer.output, keep_dims=True)
```
waleedka, line 2389: This change is unnecessary. `keep_dims` is deprecated.

mrcnn/model.py
```diff
@@ -2195,7 +2574,7 @@ def compile(self, learning_rate, momentum):
             layer = self.keras_model.get_layer(name)
             self.keras_model.metrics_names.append(name)
             loss = (
-                tf.reduce_mean(layer.output, keepdims=True)
+                tf.reduce_mean(layer.output, keep_dims=True)
```
waleedka, line 2413: This change is unnecessary. keep_dims is deprecated.

mrcnn/model.py
```diff
@@ -2796,19 +3179,55 @@ def parse_image_meta_graph(meta):
...
+def resnet_unmold_image(normalized_images, config):
+    """Takes a image normalized with resnet_mold() and returns the original."""
     return (normalized_images + config.MEAN_PIXEL).astype(np.uint8)
 
 
+def mobilenet_mold_image(images, config):
+    """Expects an RGB image (or array of images) and normalizes by
+    dividing by the median pixel value and subtracting 1.
+    """
+    return images.astype(np.float32)/127.5 - 1.0
```
waleedka, line 3037: No need to create a separate function. Both are normalizing the image, just with different values. Update MEAN_PIXEL, and add config.NORMALIZE_PIXEL such that you can control the normalization in configuration without adding new functions.

mrcnn/model.py
```diff
@@ -2796,19 +3179,55 @@ def parse_image_meta_graph(meta):
...
+
+
+def mobilenet_unmold_image(normalized_images, config):
+    """Takes an image normalized with mobilenet_mold() and returns the original."""
+    return ((normalized_images + 1)*127.5).astype(np.uint8)
```
waleedka, line 3042: Same as the comment about `mold_image()`

samples/balloon/balloon.py
```diff
@@ -300,6 +300,10 @@ def detect_and_color_splash(model, image_path=None, video_path=None):
     parser.add_argument('--video', required=False,
                         metavar="path or URL to video",
                         help='Video to apply the color splash effect on')
+    parser.add_argument('--backbone', required=False,
+                        default="resnet50",
+                        metavar="<backbone>",
+                        help='Feature Pyramid Network backbone type')
```
waleedka, line 306: Selecting a backbone should be done in the Config class like the other network architecture details.

samples/coco/coco.py
```diff
@@ -25,6 +25,13 @@
 
     # Run COCO evaluatoin on the last model you trained
     python3 coco.py evaluate --dataset=/path/to/coco/ --model=last
+
+    # Train a MobileNet224v1 model starting from ImageNet weights
+    python3 coco.py train --dataset=/path/to/coco/ --model=imagenet --backbone=mobilenet224v1
+
+    # Continue training the last MobileNet224v1 model you trained
+    python3 coco.py train --dataset=/path/to/coco/ --model=last --backbone=mobilenet224v1
```
waleedka, line 33: This is not needed once `---backbone` is removed

samples/coco/coco.py
```diff
@@ -428,13 +435,23 @@ def evaluate_coco(model, dataset, coco, eval_type="bbox", limit=0, image_ids=Non
...
                         type=bool)
+    parser.add_argument('--backbone', required=False,
+                        default="resnet50",
+                        metavar="<backbone>",
+                        help='Feature Pyramid Network backbone type')
+    parser.add_argument('--train_bn', required=False,
+                        default=False,
+                        metavar="<True|False>",
+                        help='Train batch norm layers (default=False)',
+                        type=bool)
```
waleedka, line 448: I don't see the need for a specific command line parameter to control BN training. All the other training parameters are set in Config, so I think this belongs there as well.

waleedka (changes requested):

waleedka:
@Cpruce Thanks! In addition to the inline comments above, how about this? It might be a big request, so let me know if it's not doable. Do you think you can use MobileNetV2 instead? The changes should be minimal, but the gains pretty big. V1 is a bit old by now.

mrcnn/model.py
Cpruce: Cool I'll remove this from the commit

mrcnn/model.py
Cpruce: Same as above

mrcnn/model.py
Cpruce: In fact, @JonathanCMitchell 's MobileNetV2 [implementation](https://github.com/keras-team/keras-applications/blob/master/keras_applications/mobilenet_v2.py) uses `imagenet_utils.preprocess_input` like the resnet50 implementation so I don't think there will be any change from the resnet mold/unmold.

samples/balloon/balloon.py
Cpruce: Sure I can remove this. Just thought that a command line argument would provide convenience

samples/coco/coco.py
Cpruce: Cool I'll remove it

waleedka:
@Cpruce Yes, I believe taking code from Keras-applications is fine since they use MIT license as well. You'll just need to copy their copyright statement and include it in your new file, and clarify which parts were taken from where.

PhanDuc:
@Cpruce , @chrispolo  I got this error when load the MobileNet

```
  File "boxes.py", line 529, in <module>
    "mrcnn_bbox", "mrcnn_mask"])
  File "/var/www/food-ir/boxes_segmentation/Mobile_Mask_RCNN/Mask_RCNN/mrcnn/model.py", line 2502, in load_weights
    saving.load_weights_from_hdf5_group_by_name(f, layers)
  File "/usr/local/lib/python3.5/dist-packages/keras/engine/topology.py", line 3153, in load_weights_from_hdf5_group_by_name
    ' element(s).')
ValueError: Layer #2 (named "conv1") expects 1 weight(s), but the saved weights have 2 element(s).

```

I switched to different keras versions 2.08,  2.2.4 but still got that error. Do you have any idea how to fix it?

NorthLatitudeOne:
> @Cpruce , @chrispolo I got this error when load the MobileNet
> 
> ```
> Traceback (most recent call last):
>   File "boxes.py", line 529, in <module>
>     "mrcnn_bbox", "mrcnn_mask"])
>   File "/var/www/food-ir/boxes_segmentation/Mobile_Mask_RCNN/Mask_RCNN/mrcnn/model.py", line 2502, in load_weights
>     saving.load_weights_from_hdf5_group_by_name(f, layers)
>   File "/usr/local/lib/python3.5/dist-packages/keras/engine/topology.py", line 3153, in load_weights_from_hdf5_group_by_name
>     ' element(s).')
> ValueError: Layer #2 (named "conv1") expects 1 weight(s), but the saved weights have 2 element(s).
> ```
> I switched to different keras versions 2.08, 2.2.4 but still got that error. Do you have any idea how to fix it?

Do you modified the config.py:  BACKBONE = "mobilenet224v1"

Cpruce:
@PhanDuc @NorthLatitudeOne where are you getting your weights from? Did you try --model=imagenet? If that works, converge until your heart is content and then use the best of the weights in your logs

PhanDuc:
@Cpruce , @NorthLatitudeOne  It is working when I switched `--weights `from `coco` to `imagenet`. 

Summary, before training I modified the config.py: BACKBONE = "mobilenet224v1" and use ballon as an example with my dataset.
This code below didn't work:
`python3 custom_dataset.py train --dataset=my_dataset/ --weights=coco`
And this worked:
`python3 custom_dataset.py train --dataset=my_dataset/ --weights=imagenet`

1.  I dont understand why `coco` didn't work ???
2. How to use MobileNetV2 ?

NorthLatitudeOne:
@Cpruce, @PhanDuc, Strange thing, I got the different status:
 I modified the config.py: BACKBONE = "mobilenet224v1",  and use ballon as an example with my dataset same as  Phan, 
The more I modified:
COCO_WEIGHTS_PATH = os.path.join(ROOT_DIR, "mobilenet_1_0_224_tf_no_top.h5")
and
model.train(dataset_train, dataset_val,
                 learning_rate=config.LEARNING_RATE,
                 epochs=90,
                 layers='11M+',
                 augmentation=None)
I can work well as below script because I download the mobilenet_1_0_224_tf_no_top.h5 already
python wzqcs4.py train --dataset=dataset --weights=coco

Cpruce:
@PhanDuc 
1. Are you getting your weights from the releases? Those are for resnet. I only download mobilenet weights that were trained on imagenet. I plan to release the mobilenet pretrained weights once this is merged.
2. I will be changing the Mobilenet V1 to V2 when I get some free time. 

@NorthLatitudeOne 
Sorry, I'm not sure if I understand what you're trying to say. Did it work?

NorthLatitudeOne:
@Cpruce, Sorry I have one question after training：
1、python wzqcs5.py train --dataset=dataset --weights=imagenet， training successful
2、python wzqcs5.py splash --weights=mask_rcnn_object_0120.h5 --image=190307-152053.jpg， failed
 error message:
`Running on 190307-152053.jpg
Processing 1 images
image                    shape: (1801, 3201)          min:    2.00000  max:  255.00000  uint8
  File "E:\Anaconda\lib\site-packages\numpy\lib\arraypad.py", line 935, in _normalize_shape
    shape_arr = np.broadcast_to(shape_arr, (ndims, 2))
  File "E:\Anaconda\lib\site-packages\numpy\lib\stride_tricks.py", line 176, in broadcast_to
    return _broadcast_to(array, shape, subok=subok, readonly=True)
  File "E:\Anaconda\lib\site-packages\numpy\lib\stride_tricks.py", line 128, in _broadcast_to
    op_flags=[op_flag], itershape=shape, order='C')
ValueError: operands could not be broadcast together with remapped shapes [original->remapped]: (3,2) and requested shape (2,2)

During handling of the above exception, another exception occurred:

  File "wzqcs5.py", line 446, in <module>
    video_path=args.video, image_name=args.image)
  File "wzqcs5.py", line 295, in detect_and_color_splash
    r = model.detect([image], verbose=1)[0]
  File "E:\Anaconda\lib\site-packages\mask_rcnn-2.1-py3.6.egg\mrcnn\model.py", line 2840, in detect
    molded_images, image_metas, windows = self.mold_inputs(images)
  File "E:\Anaconda\lib\site-packages\mask_rcnn-2.1-py3.6.egg\mrcnn\model.py", line 2742, in mold_inputs
    mode=self.config.IMAGE_RESIZE_MODE)
  File "E:\Anaconda\lib\site-packages\mask_rcnn-2.1-py3.6.egg\mrcnn\utils.py", line 457, in resize_image
    image = np.pad(image, padding, mode='constant', constant_values=0)
  File "E:\Anaconda\lib\site-packages\numpy\lib\arraypad.py", line 1200, in pad
    pad_width = _validate_lengths(narray, pad_width)
  File "E:\Anaconda\lib\site-packages\numpy\lib\arraypad.py", line 979, in _validate_lengths
    normshp = _normalize_shape(narray, number_elements)
  File "E:\Anaconda\lib\site-packages\numpy\lib\arraypad.py", line 938, in _normalize_shape
    raise ValueError(fmt % (shape,))
ValueError: Unable to create correctly shaped tuple from [(224, 224), (0, 0), (0, 0)]`

If I modified the utils.py as below:
`#padding = [(top_pad, bottom_pad), (left_pad, right_pad), (0, 0)]
        padding = [(top_pad, bottom_pad), (left_pad, right_pad)]  `

error message:
`Running on 190307-152053.jpg
Processing 1 images
image                    shape: (1801, 3201)          min:    2.00000  max:  255.00000  uint8
molded_images            shape: (1, 1024, 1024)       min:   -1.00000  max:    1.00000  float32
image_metas              shape: (1, 22)               min:    0.00000  max: 3201.00000  float64
anchors                  shape: (1, 261888, 4)        min:   -0.35390  max:    1.29134  float32
  File "wzqcs5.py", line 446, in <module>
    video_path=args.video, image_name=args.image)
  File "wzqcs5.py", line 295, in detect_and_color_splash
    r = model.detect([image], verbose=1)[0]
  File "E:\Anaconda\lib\site-packages\mask_rcnn-2.1-py3.6.egg\mrcnn\model.py", line 2861, in detect
    self.keras_model.predict([molded_images, image_metas, anchors], verbose=0)
  File "E:\Anaconda\lib\site-packages\keras\engine\training.py", line 1149, in predict
    x, _, _ = self._standardize_user_data(x)
  File "E:\Anaconda\lib\site-packages\keras\engine\training.py", line 751, in _standardize_user_data
    exception_prefix='input')
  File "E:\Anaconda\lib\site-packages\keras\engine\training_utils.py", line 128, in standardize_input_data
    'with shape ' + str(data_shape))
ValueError: Error when checking input: expected input_image to have 4 dimensions, but got array with shape (1, 1024, 1024)`

I don't know how to do next, could you give me any tips to solve it?

ApoorvaSuresh:
Hi,

I am using mobilenet backbone, but getting this error while inference:

ValueError: Layer #1 (named "conv1") expects 2 weight(s), but the saved weights have 1 element(s).


Any idea why?

RizanPSTU:
How to train for my own data with only one class?

Cpruce:
@NorthLatitudeOne I wouldn't modify the padding. Are you able to run splash with a resnet50 backbone? If so, what should be the shape of the tensor that is failing? How is that different in the mobilenet backbone run?

@Apsu95 Did you train from scratch on imagenet to produce the mobilenet backbone weights?

@RizanPSTU Does the splash example work for you?

Cpruce:
@waleedka can I push the MobilenetV1 backbone? The MobilenetV2 backbone is not as stable and I need to dedicate time that I do not have right now to debug. Once things settle down and I resolve the convergence issues with MobilenetV2 as a backbone, I can submit another PR.

zhoujinhai:
> @PhanDuc
> 
> 1. Are you getting your weights from the releases? Those are for resnet. I only download mobilenet weights that were trained on imagenet. I plan to release the mobilenet pretrained weights once this is merged.
> 2. I will be changing the Mobilenet V1 to V2 when I get some free time.
> 
> @NorthLatitudeOne
> Sorry, I'm not sure if I understand what you're trying to say. Did it work?

hello, Do you have the pretrained weights of mobilenetv2 mask rcnn?

PhanDuc:
@zhoujinhai , sorry I don't have it.

zhoujinhai:
Ok thanks


发自我的iPhone

------------------ Original ------------------

zhoujinhai:
> @waleedka can I push the MobilenetV1 backbone? The MobilenetV2 backbone is not as stable and I need to dedicate time that I do not have right now to debug. Once things settle down and I resolve the convergence issues with MobilenetV2 as a backbone, I can submit another PR.

@Cpruce I have training my data by mobileNetV2 backbonek, but the detect result is bad, Do you have the pretrained weights of mobilenetv2 mask rcnn?

suchiz:
Hi @Cpruce, you are the only one I found who succeeded to deploy it on Android. I have a few questions if you don't mind. Was if tensorflow-android or tflite ? if it was tensorflow-android, you ran it from the frozen-graph only right, wasn't it slow ? If it is tflite, how did you manage to convert custom layers such as BatchNorm, Proposal Layers etc...

Thank you in advance ...

Cpruce:
@zhoujinhai Sorry, I do have pretrained weights for mobilenetV2, but as I said earlier the training did not converge to an acceptable point. I believe further architecture changes will need to be made to make pixel-level classifications more applicable with mobilenetV2's representational capacity. IMHO, skip mobilenetV2 and use efficientNet or MNASNet.

@suchiz It was tensorflow-android since too many of the operations were not implemented in tflite and yes, it was very slow.

JonathanCMitchell:
I have the pretrained weights for general MobileNetV2 (without Mask-RCNN) on my repo here https://github.com/JonathanCMitchell/mobilenet_v2_keras/releases which you can use to train with MRCNN, but you should keep in mind that MobileNetV2 is not great for MRCNN and you should probably use EfficientNet like Cory said.

adsk2050:
> Hi,
> 
> I am using mobilenet backbone, but getting this error while inference:
> 
> ValueError: Layer #1 (named "conv1") expects 2 weight(s), but the saved weights have 1 element(s).
> 
> Any idea why?

@ApoorvaSuresh did you find out what the issue was? I am getting a similar error: 
```
ValueError: Layer #362 (named "anchors") expects 1 weight(s), but the saved weights have 0 element(s).
```

Tomta10:
I had a similar problem, I got this weird error when adding model weights, can anyone help me? thanks
ValueError: Layer #100 (named "anchors") expects 1 weight(s), but the saved weights have 0 element(s).

zhoujinhai:
您的邮件已收到 谢谢

NikeeShrestha:
> I had a similar problem, I got this weird error when adding model weights, can anyone help me? thanks ValueError: Layer #100 (named "anchors") expects 1 weight(s), but the saved weights have 0 element(s).

Did you solve this problem? I am having same error.

suso86:
Did someone resolve the problem: Layer #362 (named "anchors") expects 1 weight(s), but the saved weights have 0 element(s).?

travishsu:
I replace the [L2139](https://github.com/Cpruce/Mask_RCNN/blob/master/mrcnn/model.py#L2139):
```python
anchors = KL.Lambda(lambda x: tf.Variable(anchors), name="anchors")(input_image)
```
with
```python
anchors = tf.constant(anchors, name="anchors")
```

suso86:
> 

thank so much. I have aplicated this and it solved. Do you know if This repository (https://github.com/Cpruce/Mask_RCNN) function with tf 2.0?? I´m working with google colab and I have to work with tensorflow 2.0

simpo-lei:
> Did someone resolve the problem: Layer #362 (named "anchors") expects 1 weight(s), but the saved weights have 0 element(s).?

Same problem here. Looks like not too many people work on re-training.

PR #130, opened 2017-12-11
refactor & add CI

Borda:
make the code as a package and add a CI testing

PR #391, opened 2018-04-03
make a package

Borda:
refactoring structure, initially in #130 

* make the code as a package
* add requirements
* add setup


Borda:
@waleedka @hadim ^^

hadim:
`pip install --no-deps https://github.com/Borda/Mask_RCNN/archive/refactor.zip` works like a charm. Thanks.

README.md
```diff
@@ -161,10 +161,19 @@ You can also [join our team](https://matterport.com/careers/) and help us build
...
+All required packages are listed in standard file:
+```bash
+pip3 install -r requirements.txt
+```
+
+## Installation
+
+The installation as possible using following command:
+```bash
+python3 setup.py build_ext --inplace
```
waleedka: The `build_ext` is for extensions, right? Do we need this?

mrcnn/shapes.py
```diff
@@ -14,8 +14,8 @@
 import numpy as np
```
waleedka: This file is part of the shapes sample. Should be in the `samples` folder.

samples/balloon/inspect_balloon_data.ipynb
```diff
@@ -49,13 +49,13 @@
     "\n",
     "# Import Mask RCNN\n",
     "sys.path.append(ROOT_DIR)\n",
-    "import utils\n",
-    "import visualize\n",
-    "from visualize import display_images\n",
-    "import model as modellib\n",
-    "from model import log\n",
+    "import mrcnn.utils\n",
```
waleedka: I believe this should be `from mrcnn import utils`, otherwise references to `utils` in the code would fail. The same applies to several other import statements in this file and others.

mrcnn/model.py
```diff
@@ -2036,7 +2030,7 @@ def find_last(self):
         dir_name = os.path.join(self.model_dir, dir_names[-1])
         # Find the last checkpoint
         checkpoints = next(os.walk(dir_name))[2]
-        checkpoints = filter(lambda f: f.startswith("mask_rcnn"), checkpoints)
+        checkpoints = filter(lambda f: f.startswith("mrcnn"), checkpoints)
```
waleedka: This would cause it to fail to find previously saved checkpoints.

mrcnn/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+
+import matplotlib
+import numpy as np
+
+# in case you are running on machine without display, e.g. server
+if os.environ.get('DISPLAY', '') == '' \
+        and matplotlib.rcParams['backend'] != 'agg':
+    logging.warning('No display found. Using non-interactive Agg backend')
+    # https://matplotlib.org/faq/usage_faq.html
+    matplotlib.use('Agg')
```
waleedka: I could be wrong, but I think matplotlib does its best to handle different environments automatically. In which case is this needed? Also, if it is needed then let's have it in `visualize.py` or somewhere else more visible in case someone needs to edit it for their environment. Maybe in a separate PR so we don't cause delay to this one.

mrcnn/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+
+# parse the numpy versions
+np_version = [int(i) for i in np.version.full_version.split('.')]
+# comparing strings does not work for version lower 1.10
+if np_version >= [1, 14]:
+    np.set_printoptions(legacy='1.13')
```
waleedka: Why force using a legacy setting? Also, maybe it should be in a separate PR as it's not related to this update.

waleedka:
@Borda I added a few comments above. Thank you so much for this PR. 

Other than the above:
- Since you already started the samples reorganization that I was planning to do, may I suggest the following so each sample has its own folder:
  * demo.ipynb stays in root. It's an intro to the repository.
  * shapes.py and train_shapes.ipynb -> samples/shapes/
  * coco.py, inspect_model, inspect_data, inspect_weights -> samples/coco/
- I'm guessing you used `mrcnn` rather than `mask_rcnn` because it's shorter? Any other reason? I'm inclined to use mask_rcnn for clarity when people search for the package. At least, naming the library as mask_rcnn on PyPi.

mrcnn/__init__.py
Borda: This is just preventing crash visualization in figures if you run it on a server without any screen or session (also some CI have similar characteristic)

mrcnn/__init__.py
Borda: It is mainly for testing and using doctest the requirement doe not specify an exact version and when you run test locally you may get different output than running it on CI

Borda:
I used mrcnn so it does not confuse with the repository name and also on PyPi you can use the mask_rcnn as package name (e.g. skimage is installed from pip via scikit-image)

Borda:
@waleedka ^^

waleedka:
Working on it as we speak :)

Borda:
@waleedka may you merge it before some other PR because than I always have to merge it and check for collision and the last one was not so simple I did not succeed to rebase it... or let me know before "merge" and I will try one last merge/rebase... :)

waleedka:
Also, the Jupyter notebooks are now failing with this error:

```
WARNING:root:No display found. Using non-interactive Agg backend
/deeplearning/mask_rcnn/mrcnn/visualize.py:22: UserWarning: 
This call to matplotlib.use() has no effect because the backend has already
been chosen; matplotlib.use() must be called *before* pylab, matplotlib.pyplot,
or matplotlib.backends is imported for the first time.
```

Borda:
is the Jupiter really failing or just printing this warning?

Borda:
@waleedka as the package name after install is the same as the package in the folder you really do not need to install anything using setup and will run from the local package... is it what you asked?

waleedka:
1. I'm trying to run the samples and the notebook *without* running the setup command. Because that's what most people who are already using the project would likely be doing. This gives me this error:
```
  File "coco.py", line 49, in <module>
    from mrcnn.config import Config
ImportError: No module named 'mrcnn'
```


2. You're right, Jupyter notebook is not failing, but it's a pretty big error. 

image

Borda:
Add 1) you need to run the ipython in the root OR I can add path to the root it beginning of each ipynb
Add 2) I agree that it is not nice, so I will remove the matplotlib.use()

waleedka:
Notice that ROOT_DIR is used in the notebooks, so removing it would break them

hadim:
About the import thing you should do the following:

`
Import os
os.path.append("../mrcnn")
`

Borda:
@hadim I added the path to the root using the relevant number of ".."

hadim:
That what I use for my Python code inside the same repo of a Python libray.

So users have the choice:

- Clone and import the local library.
- pip install it.

waleedka:
Merged, finally! 

I rebased the branch and made a few updates and fixes. I did it outside of github because it was too complex, so don't worry about the message above that says there are conflicts. It's merged correctly and everything looks good. 

@Borda Thank you so much for making this happen.

PR #448, opened 2018-04-14
loss weights

Scitator:
Hi!
I found this implementation quite useful for playing around with object detection and instance segmentation.

Nevertheless, for the object detection training (no mask required) I haven't found a way to exclude the masks from training pipeline. By this way, I create a small hack and multiply segmentation loss by 0.

To get something better, I decide to use additional loss weights in model config. So, here is PR.


waleedka:
Thanks!

I've been considering adding a config flag to let the model work in bounding box only mode, along with a `load_bbox()` method in the dataset class to be called when that mode is enabled. Haven't gotten around to doing it though.

PR #487, opened 2018-04-26
Update utils.py

yezqNLP:
if the model detect nothing,the return r['mask'].shape=[0,28,28],and the following np.reshape（） would report an error, so I add two lines.


waleedka:
Thank you!

PR #515, opened 2018-05-04
Working with py2

lqhuy93:
this PR makes Mask_RCNN working with python 2 by adding some float casting.

To run the demo, I suggest you guys download `COCO_MODEL_URL = "https://github.com/matterport/Mask_RCNN/releases/download/v2.0/mask_rcnn_coco.h5"` manually since I still see some error in this line https://github.com/matterport/Mask_RCNN/blob/master/mrcnn/utils.py#L852


akshaytrivedi:
Thanks a lot!! Works for me

HectorAnadon:
Can you give some instructions about how to use it? Thanks!

waleedka:
May I suggest using `from __future__ import division` as described here.

https://stackoverflow.com/questions/21316968/division-in-python-2-7-and-3-3

It's likely a simpler solution and catches all cases so we don't have to chase them one by one.

upwindflys:
Thanks a lot!

musematics:
Thank you very much. This perfectly solved my problem. I must pay more attention to the version of python.

PR #516, opened 2018-05-04
Mobilenet V1 and V2 as Backbone + py2/3 compability + several other things

gustavz:
I am pretty sure it wont merge easily as i changed a lot of stuff.
But this fork works both on python 2 and 3 and it includes mobilenet v1 and v2 aswell as the default resnet50 and resnet101.
Furthermore i added some other features, like the support for coco to pass class_names instead of class_ids as this feels more natural to me.
There are also some updates to image molding and some speed hacks that are concerned in other issues.

Maybe its work a look.


Zumbalamambo:
@GustavZ  resource is getting exhausted while trying to train using your code. Though i have reduced batch size and the image resolution. Also your code doesnt work with balloon example

asimay:
> @GustavZ There is a mobile_mask_rcnn_coco.h5 in your repo, is it the weight of MobileNet V1 or V2? Thank you.

it's V1 version

asimay:
@GustavZ ,  hi, dear, I have a question:

why https://github.com/matterport/Mask_RCNN/blob/master/mrcnn/model.py#L2188

loss is defined to multiply by length of output?? it's weired.. Could you please explain it? thank you very much!

cvtower:
Hi all,
Seems like the mobilenet v2 implement is not complete(at least mbv2 imagenet pretrained model not loaded as mobilenetv1), i guess no solid pytorch pretrained model ever released then.

Hanson13:
@gustavz Thanks for your code, it really helps me! I used the code to train on my own dataset and got a trained weight successfully. However, when I used the weight to inference on my dataset, the results only contained the bounding box of detection but **no mask of segmentation**. Does anyone have any idea why it happened? Thanks a lot!

Best regards,
Hanson

zhoujinhai:
> @gustavz Thanks for your code, it really helps me! I used the code to train on my own dataset and got a trained weight successfully. However, when I used the weight to inference on my dataset, the results only contained the bounding box of detection but **no mask of segmentation**. Does anyone have any idea why it happened? Thanks a lot!
> 
> Best regards,
> Hanson

@Hanson13 Do you have try mobileNetv2 backbone? I used [code](https://github.com/Cpruce/Mask_RCNN) to train, but the detect result is not good, if you use mobileNetV2, can you provide the pretrained weigths?

zhoujinhai:
@gustavz @asimay Do you have the pretrained weigths of mobilenetv2 as backbone?

RiyazAina-DeepML:
> I am pretty sure it wont merge easily as i changed a lot of stuff.
> But this fork works both on python 2 and 3 and it includes mobilenet v1 and v2 aswell as the default resnet50 and resnet101.
> Furthermore i added some other features, like the support for coco to pass class_names instead of class_ids as this feels more natural to me.
> There are also some updates to image molding and some speed hacks that are concerned in other issues.
> 
> Maybe its work a look.

thanks alot . i have one question , can we train this mobile_mask_rcnn with original via data format?
when i tried this mmrcnn on via data format , i am not getting any detection as well as mask prediction .
what will be the issue can u help

PR #541, opened 2018-05-09
Fix ROOT_DIR path and autodownload arg parse

JonathanCMitchell:
This commit changes the ROOT_DIR from `../../` for the current working directory, which is the intended use of it after all. Therefore, the `DEFAULT_LOGS_DIR` will be placed under the project directory root, and not in `$HOME/logs` when the `--logs` argument is not provided in command line execution.

Now to call training, you need so simply supply the `--autodownload` flag. Before, the flag was passed in as a string and was automatically set to true because when you bool(<string>) it is always true. In this case, we can interpret the autodownload flag as a boolean directly when it is supplied.

Additionally, we no longer need to explicitly set the `PYTHONPATH` from command line when we train, and the current command line expressions that are defined in the readme still work, we simply removed a step and provide additional clarity on which directory will always be the project root.

### DIFF
--- a/samples/coco/coco.py
+++ b/samples/coco/coco.py
@@ -48,7 +48,7 @@ import urllib.request
 import shutil
 
 # Root directory of the project
-ROOT_DIR = os.path.abspath("../../")
+ROOT_DIR = os.getcwd()
 
 # Import Mask RCNN
 sys.path.append(ROOT_DIR)  # To find local version of the library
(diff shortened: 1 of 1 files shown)


waleedka:
Setting the root directory could be tricky. For me, the way it's setup now works. To run `coco.py`, I usually go to the coco folder and run it from there:
```
cd ~/mask_rcnn/samples/coco
python3 coco.py ....
```
So my working directory is `~/mask_rcnn/samples/coco`, and `../../` sets DOOR_DIR to the root of the project, which is how it should be. Is that not how it works for you?

I agree with the ``--download`` change. But not with removing the `help` parameter.

And, finally, I didn't understand your comment about `PYTHONPATH`. I don't set that when I run the code.

Issue #544, opened 2018-05-10
Bug or any deeper consideration?  The callbacks in "/mrcnn/model.py"


ziyigogogo:
It seems that the "log_dir" is created in the callback "keras.callbacks.TensorBoard".

But in my application, I don't need the "tensorboard" analysis which could be a really time consuming job for my hardware condition.

So, i comment the 2298-2299 lines in file "/mrcnn/model.py".

```python
keras.callbacks.TensorBoard(log_dir=self.log_dir,
                                        histogram_freq=0, write_graph=True, write_images=False),
```

After that an error happened:
OSError: Unable to create file (unable to open file: name = 'D:\XXXX\mask_rcnn\logs\XXXX20180510T1347\mask_rcnn_XXXX_0001.h5', errno = 2, error message = 'No such file or directory', flags = 13, o_flags = 302)

A workaround now is to add the following codes after line ~~2295~~ 2237 in "model.py":

```python
if not os.path.exists(self.log_dir):
    os.makedirs(self.log_dir)
```

But i am not sure that if there is a deeper consideration for the current behave, please kindly let me know if so@waleedka

ziyigogogo:
Hmm, i think do the creation @line 2237 in the "set_log_dir" function is better

waleedka:
Thanks you! The PR is merged so I'm closing this.

PR #545, opened 2018-05-10
Bug fix: create log_dir

ziyigogogo:
As I described in [issue#544](https://github.com/matterport/Mask_RCNN/issues/544),

My very personal opinion is that : "**Keras.callbacks**" should simply behave in a "**Plug&Play**" style.
Thus it is natural that if we unplug a "callback", the rest of the system should run without any problem.

The current codebase( I have to say it's great and amazing already) did not consider the situation when people do not use "TensorBoard", so that the log dir would not be created(it's now created by the "**tf.summary.FileWriter**" inside "**keras.callbacks.TensorBoard**") and would result in the "Directory not found" error when saving "**ModelCheckpoint**"

And the solution is pretty straight forward, just check the "**log_dir**" and create it if not exists

### DIFF
--- a/mrcnn/model.py
+++ b/mrcnn/model.py
@@ -2235,6 +2235,10 @@ class MaskRCNN():
         self.log_dir = os.path.join(self.model_dir, "{}{:%Y%m%dT%H%M}".format(
             self.config.NAME.lower(), now))
 
+        # Create log_dir if not exists
+        if not os.path.exists(self.log_dir):
+            os.makedirs(self.log_dir)
+
         # Path to save after each epoch. Include placeholders that get filled by Keras.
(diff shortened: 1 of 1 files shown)


waleedka:
Agreed! The Keras callbacks should be optional. Thanks.

ziyigogogo:
@waleedka 
hi, there. This PR is my very first fix last month. Since I found some other problems related to log_dir here these days when playing around this project. I will update a PR ASAP today.
Two main problems of the current version:
1. we should not create log_dir when "dectect".

2. Since the method "set_log_dir" is used at two function: "__init__" and "load_weights", sometimes we lunch the training at the end of a minute let's say 201806061530, two folders: "201806061530" and "201806061531" may be created.

PR #607, opened 2018-05-29
Not load model weights when training from scratch

viirya:
When going to train coco model from scratch by specifying `model = last`, loading model weights will cause error. We should skip weight loading if we can't find any last model.

### DIFF
--- a/samples/coco/coco.py
+++ b/samples/coco/coco.py
@@ -469,9 +469,10 @@ if __name__ == '__main__':
     else:
         model_path = args.model
 
-    # Load weights
-    print("Loading weights ", model_path)
-    model.load_weights(model_path, by_name=True)
+    # Load weights if model_path is defined.
+    if model_path is not None:
+        print("Loading weights ", model_path)
+        model.load_weights(model_path, by_name=True)
 
     # Train or evaluate
     if args.command == "train":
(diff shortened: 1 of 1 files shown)


waleedka:
If I want to train from scratch I don't pass the `--model` argument.

On the other hand, if I pass `--model=last` then I intend for the training to continue from where I left off. If the last weights can't be found, then I prefer the code tells me about the problem rather than do something else different from what I wanted (i.e. train from scratch).

viirya:
Ok. That makes sense. Then I think it is better to display some useful message.

waleedka:
Good point. Done here https://github.com/matterport/Mask_RCNN/commit/cbff80f3e3f653a9eeee43d0d383a0385aba546b

waleedka:
Huh! I just noticed that you pushed a commit to this PR. I guess we were working in parallel.

I did the update in `model.py` so it captures this case for all the samples, not just the COCO sample.

viirya:
Ok. Looks good. Thanks.

enoceanwei:
@waleedka,

 Hi, Sorry to disturb you, I am suffering a problem. 

I had changed the CNN backbone structure, so I need to train a model from scratch without using the pre-train weights from coco h5. When I train the new model, I just comment out the 'model.loadweight' this line in the codes. 

However, the result is terrible, the new model (just train for the head layers) can not detect anything in my own datasets almost, On the compare, if I use the pretrain weights, the result is pretty good, I have no idea about it, could you give me some suggestion? many thanks.

Kind regards

Wei

PR #613, opened 2018-05-30
Few dependencies fixes, and bug fixes.

pvrsa:
Urllib requests package is deprecated. Hence changed it to urllib.retreive().

Fixed norm_boxes() in utils.py , which was returning a zero-matrix on divide. So used asfloat(np.float32)

### DIFF
--- a/mrcnn/utils.py
+++ b/mrcnn/utils.py
@@ -17,7 +17,7 @@ import scipy
 import skimage.color
 import skimage.io
 import skimage.transform
-import urllib.request
+import urllib
 import shutil
 import warnings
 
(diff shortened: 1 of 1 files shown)


waleedka:
I couldn't find anywhere that says urlopen() is deprecated. Can you point me to your resource? In fact, from what I read, `urlretrieve()` is an old port from Python 2 and might be deprecated.

As for the other fix, can you provide more details? I don't get a zero matrix result. Are you using Python 2.7 by any chance? If so, that might be the issue, and in that case I think a better fix would be to use the future import as explained in this page. I don't have Python 2.7 to test this, but if you try it and it fixes the issue then I can merge it:
https://stackoverflow.com/questions/21316968/division-in-python-2-7-and-3-3

pvrsa:
Oh yeah, I was using Python 2.7 when I ran into these issues.
I think they are not a problem when using Python 3

waleedka:
Okay, thanks. Then I'll close this for now. There is another PR https://github.com/matterport/Mask_RCNN/pull/515  trying to make the code work on 2.7.

PR #648, opened 2018-06-06
bug fix: un-excepted creation(issue#545) is now removed

ziyigogogo:
Hi @waleedka , 
I found some other problems related to my first PR these days when playing around with this repo.

Two main problems of the last version:

1. we should not create log_dir when "**dectect**".

2. Since the method "**set_log_dir**" is used at 2 places: "**\_\_init\_\_**" and "**load_weights**", sometimes we lunch the training at the end of a minute let's say 201806061530, two folders: "201806061530" and "201806061531" may be created.

This PR fixed problems described above, thank you.

### DIFF
--- a/mrcnn/model.py
+++ b/mrcnn/model.py
@@ -2251,12 +2251,12 @@ class MaskRCNN():
                 self.epoch = int(m.group(6)) - 1 + 1
                 print('Re-starting from epoch %d' % self.epoch)
 
-        # Directory for training logs
-        self.log_dir = os.path.join(self.model_dir, "{}{:%Y%m%dT%H%M}".format(
-            self.config.NAME.lower(), now))
+        # Directory for training logs:  Init self.log_dir for only once to avoid duplicate creation
+        if not hasattr(self, "log_dir"):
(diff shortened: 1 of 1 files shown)


waleedka:
I see the problem. I'm wondering, wouldn't it be easier to create the `log_dir` in the `train()` function before training starts? This solves both problems without adding addition checks and conditions.

waleedka:
> why we should call "set_log_dir()" both in "__init__" and "load_weights()" instead of only in "load_weights()"

Because calling `load_weights()` is optional. If someone trains a model from scratch without loading any weights, the logs still need to be saved in a directory.

akTwelve:
I've been looking at this problem too, since it broke my training code (and appears to break the train_shapes sample too). I think the "if not exists, create directory" code that @ziyigogogo originally added should be moved to the train() function. It is the cleanest solution with the most predictable results. Since it would put new trained weights in the same directory, it would allow users to see a clear history rather than spreading out the weights across multiple, hard to track, log directories.

I have a commit that does that in my fork of the repo here: https://github.com/akTwelve/Mask_RCNN/commit/256279c362b051a30fb6b974b873373346ecd387

waleedka:
I merged https://github.com/matterport/Mask_RCNN/pull/823 which solves the same problem. Closing this.

Issue #566, opened 2018-05-16
load_weights_from_hdf5_group_by_name unknown attribute in latest keras


castleguarders:
In the demo notebook, the line below
model.load_weights(COCO_MODEL_PATH, by_name=True)
is used to load weights, this no longer works in Keras. I tried with the latest github pull, as well as the pip default install on ubuntu 16.04. I had to revert to Keras 2.0.8 specifically to get this to work. Looks like model.load_weights and model.save_weights is the way going forward in keras.

        if by_name:
            topology.load_weights_from_hdf5_group_by_name(f, layers)
        else:
            topology.load_weights_from_hdf5_group(f, layers)

gtgalone:
I made pull request. Until it merged, You can replace topology with saving. Topology engine will be deprecated :)

PratibhaT:
I made the change still I'm getting error.

Loading weights  /home/pratibha/dl_semseg/Mask_RCNN/mask_rcnn_coco.h5
Traceback (most recent call last):
  File "balloon.py", line 353, in <module>
    "mrcnn_bbox", "mrcnn_mask"])
  File "/home/pratibha/anaconda3/lib/python3.5/site-packages/mask_rcnn-2.1-py3.5.egg/mrcnn/model.py", line 2119, in load_weights
AttributeError: module 'keras.engine.topology' has no attribute 'load_weights_from_hdf5_group_by_name'

gtgalone:
@PratibhaT After changing that, You have to rollback your ipynb file to original ipynb from this repo :)

gtgalone:
@PratibhaT After changing that, You have to rollback your ipynb file to original ipynb

tanzf:
@gtgalone You have to rollback your ipynb file to original ipynb？How to do the specific operation？

gtgalone:
@tanzf ```git checkout -- .``` type this on your terminal or remove Mask_RCNN directory and then clone this repo again

yshvrdhn:
@tanzf
After changing the model.py file don't forget to reinstall the maskrcnn package using python3 setup.py install  that shoould make it work I used the same instructions mentioned by @gtgalone

PratibhaT:
@gtgalone @yshvrdhn thank you..it worked...but i'm not able to run it on my gpu...in config.py it says that put GPU_COUNT=1 for training on cpu, So, changed that to 2, and it raised an error. Which files did you change to run training on gpu.

yshvrdhn:
[...]

PR #662, opened 2018-06-09
Fix Keras engine module topology to saving

gtgalone:
Fix for Issue #566

### DIFF
--- a/mrcnn/model.py
+++ b/mrcnn/model.py
@@ -2094,7 +2094,7 @@ class MaskRCNN():
         exlude: list of layer names to excluce
         """
         import h5py
-        from keras.engine import topology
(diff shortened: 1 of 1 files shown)


gtgalone:
Nope, But most are using Keras latest version and Keras notices that topology module is deprecated.

trancept:
Got the same problem. It seems Keras default pip3 version does not support topology any-more.
Thanks @gtgalone to point this out, it really help me.

waleedka:
Actually, this does break the code on earlier versions of Keras.
```Python
>>> keras.__version__
'2.1.6'

>>> from keras.engine import saving
ImportError: cannot import name 'saving'
```

I'll merge it, and then add a check to handle older versions.

PR #674, opened 2018-06-13
Updated Coco Example, SKImage Resize Reference Updates

Veemon:
- Updated the coco example for use with the updated 2017 dataset.
- Updated references to the skimage resize function to include the false anti-aliasing flag. This is done so to both keep the intention, yet comply with the upcoming 0.15 release.

### DIFF
--- a/mrcnn/model.py
+++ b/mrcnn/model.py
@@ -2062,8 +2062,7 @@ class MaskRCNN():
         """Finds the last checkpoint file of the last trained model in the
         model directory.
         Returns:
-            log_dir: The directory where events and weights are saved
-            checkpoint_path: the path to the last checkpoint file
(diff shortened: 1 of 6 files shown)


waleedka:
Thanks @Veemon. The COCO 2017 is great. I'm ready to merge it. 

The skimage update is accurate, but it breaks the code for anyone who has skimage earlier than 0.14. Since 0.14 was released about 6 weeks ago, it's too early to expect everyone to upgrade.

Can you remove the skimage updates from this PR? Then I'll merge it, and we can handle the skimage issue separately. I'm thinking maybe it's best to create an internal resize() function utils.py so all changes can be handled in one location, and we can add version checking there.

Veemon:
Refer to #771

PR #687, opened 2018-06-16
Updating future potentially deprecated code

theabhinavdas:
Documentation says urllib.request "might become deprecated at some point in the future".

IMHO, it's better to provide examples that do not make users rely on API that has this label.

### DIFF
--- a/mrcnn/utils.py
+++ b/mrcnn/utils.py
@@ -17,7 +17,7 @@ import scipy
 import skimage.color
 import skimage.io
 import skimage.transform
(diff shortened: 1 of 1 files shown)


waleedka:
My understanding is that `urlib.request` was newly added in Python 3, and I don't see any mention of it being depreciated. Also, `urllib.urlopen()` is removed in Python 3. So, I'm afraid I can't merge this PR because it would break the code in Python 3, which is what this repo is verified to run on.

PR #688, opened 2018-06-17
Add custom callbacks to model training

nicofirst1:
Add an optional parameter for calling a list of keras.callbacks to be add to the original list.

### DIFF
--- a/mrcnn/model.py
+++ b/mrcnn/model.py
@@ -2266,7 +2266,7 @@ class MaskRCNN():
             "*epoch*", "{epoch:04d}")
 
     def train(self, train_dataset, val_dataset, learning_rate, epochs, layers,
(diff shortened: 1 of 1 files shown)


hadim:
Is this going to be merged? That would very useful.

hadim:
ping @waleedka

waleedka:
Thanks! Merged.

There is a small issue, though, related to using mutable values as default arguments. I know because I've been bitten by that before and finding the issue took a very long time. See: 

http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

I fixed that issue in  https://github.com/matterport/Mask_RCNN/commit/30292efc95a236641ff3e0487321ef6135ff0f06

PR #740, opened 2018-07-02
replaced data_generator with a subclass of keras.utils.Sequence to avoid duplicate data when multiprocessing=True

coreyhu:
Keras is moving away from using generators to load in data due to concerns of them not being threadsafe during multiprocessing. I just moved data_generator's generator code from into DataGenerator, a subclass of keras.utils.Sequence where __getitem__(idx) returns the inputs/outputs of a new batch with the given index. Using Kera's Sequence class should fix possible data redundancy concerns when using multiprocessing


waleedka:
Thanks @coreywho. I think there are a couple of issues:

- The image_index is set to -1 on every call to `__getitem__()`, so every batch will return the same images. Also, the `idx` argument of this function is not used.
- The code that captures image loading errors and skip the image was removed. When training on a lot of images, it's common to have random errors on a few of them. It's better to log it and skip rather than stop the training.

waleedka:
@coreywho Do you happen to know if this PR fixes the multi-processing issue for Windows?

luckyssc (approved):

prafulag:
Is this code getting merged to master any soon ? 
Is it safe to use this new DataGenerator in place of data_generator in model.py ?

PR #756, opened 2018-07-06
Fixed warnings resulting from calls to skimage.transform.resize

maxfrei750:
Calls to skimage.transform.resize result in the warning "Anti-aliasing will be enabled by default in skimage ...". Therefore, the parameter anti_aliasing=True was added to each call of skimage.transform.resize.


waleedka:
See my comments about skimage resize in https://github.com/matterport/Mask_RCNN/pull/674

The other part of your PR related to the Keras fix is already merged, and the other changes in requiremetns.txt seem unrelated to this PR.

PR #792, opened 2018-07-19
bug fix: the output_shape of roi_gt_class_ids is incorrect

Viredery:
when trying to add online hard example mining, I find the output shape of roi_gt_class_ids is not correct and the code can't run.


waleedka:
Good catch! Thank you.

PR #797, opened 2018-07-19
Added automated epoch recognition for Windows.

maxfrei750:
Added a regex for Windows paths.


maxfrei750:
Found a more elegant solution (see #798).

PR #798, opened 2018-07-19
Added automated epoch recognition for Windows.

maxfrei750:
Added a regex for Windows paths.

### DIFF
--- a/mrcnn/model.py
+++ b/mrcnn/model.py
@@ -2259,8 +2259,14 @@ class MaskRCNN():
         if model_path:
             # Continue from we left of. Get epoch and date from the file name
             # A sample model path might look like:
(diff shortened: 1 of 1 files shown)


HHoofs:
Why not ~~~regex = r".*[/\\][\w-]+(\d{4})(\d{2})(\d{2})T(\d{2})(\d{2})[/\\]mask\_rcnn\_[\w-]+(\d{4})\.h5"~~~

maxfrei750:
The most likely reason may very well be that I'm not good at regular expressions. 😅 

Have you tested you solution?

HHoofs:
For me it works, but maybe some tests would be a good idea?

maxfrei750:
@HHoofs I tested your solution and it worked for me as well. Therefore I added a commit to this PR.

HHoofs:
Nice, I think this is a nice easy fix

PR #823, opened 2018-07-29
Bugfix for MaskRCNN creating empty log_dir that breaks find_last()

akTwelve:
- Current implementation creates self.log_dir in set_log_dir() function,
  which creates an empty log directory if none exists. This causes
  find_last() to fail after creating a model because it finds this new
  empty directory instead of the previous training directory.
- New implementation moves log_dir creation to the train() function to
  ensure it is only created when it will be used.


rgommers:
This looks like the right fix to me. Would be good to merge as is.

maxfrei750:
I can also confirm that his works flawlessly.

waleedka:
Merged! Thank you, @akTwelve for the fix. And everyone for confirming it.

PR #827, opened 2018-08-02
Add num_intra_threads/num_inter_threads.

pandaoxin:
Add options for num_intra_threads/num_inter_threads.

### DIFF
--- a/mrcnn/config.py
+++ b/mrcnn/config.py
@@ -203,6 +203,11 @@ class Config(object):
     # Gradient norm clipping
     GRADIENT_CLIP_NORM = 5.0
 
+    # For num_intra_threads/num_inter_threads.
+    # Use default settings if no change for value 0.
+    NUM_INTRA = 0
+    NUM_INTER = 0
+
     def __init__(self):
         """Set values of computed attributes."""
         # Effective batch size
--- a/mrcnn/model.py
+++ b/mrcnn/model.py
@@ -1843,6 +1843,11 @@ class MaskRCNN():
         self.model_dir = model_dir
         self.set_log_dir()
         self.keras_model = self.build(mode=mode, config=config)
+        if config.NUM_INTRA > 0 and config.NUM_INTER > 0:
+            session_config = tf.ConfigProto(intra_op_parallelism_threads=config.NUM_INTRA,\
+                     inter_op_parallelism_threads=config.NUM_INTER)
+            session = tf.Session(config=session_config)
+            K.set_session(session)
 
     def build(self, mode, config):
         """Build Mask R-CNN architecture.
--- a/samples/coco/coco.py
+++ b/samples/coco/coco.py
@@ -405,6 +405,14 @@ if __name__ == '__main__':
     parser.add_argument("command",
                         metavar="<command>",
                         help="'train' or 'evaluate' on MS COCO")
+    parser.add_argument("--num_intra_threads", required=False,
+                        default=56,
+                        metavar="<num intra threads>",
+                        help="Num intra threads (default=56)")
+    parser.add_argument("--num_inter_threads", required=False,
+                        default=1,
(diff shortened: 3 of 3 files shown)


rgommers:
@pandaoxin can you add a bit of context to this PR (explain what it does and why it's important)? Also, how does it interact with the exsiting `multiprocessing` use?

pandaoxin:
@rgommers It 's about Eigen thread pool settings.
For intra_op_parallelism_threads: Nodes that can use multiple threads to parallelize their execution will schedule the individual pieces into this pool.
For inter_op_parallelism_threads: All ready nodes are scheduled in this pool.
And users could tune both of them to optimize CPU performance.

rgommers:
Thanks, that helps. I do think you need to deal with the interaction with `multiprocessing`, which is by default on and uses all CPUs already. See [here](https://github.com/matterport/Mask_RCNN/blob/1ac88be880f46ec6a0d22fbae2bbd8d829dfa1ef/mrcnn/model.py#L2369). Having both multiple threads and multiple processes will not interact nicely I'd think.

pandaoxin:
@rgommers On my machine, multiprocessing.cpu_count() is 112, meaning 112 processes. While default Eigen num_intra_threads is 56, meaning 56 threads for each process. If we add the option to tune Eigen thread number, tuning for performance would be more friendly.

rgommers:
Ah yes, so you already ran into the problem I'm talking about:) In general, users don't want multiprocessing + multi-threading I think; that will just result in lots of unnecessary overhead. So I'm suggesting to at least document how this works for `mrcnn` with this new setting. Also, it would probably be very helpful to make the number of workers configurable as well (tuning threads when `n_processes == cpu_count` isn't going to do too much ...).

I just had to do something similar: https://github.com/rgommers/Mask_RCNN/commit/2fe7aa251017723111926f67bc0400648231d4d0. Adding a setting for `workers` too with a recommendation of a good balance between threads/processes would be nice.

pandaoxin:
@rgommers Great advice. I have added some comments to document the usage of num_intra/inter_threads. For option of workers, may I raise it in another PR?

pandaoxin:
@dcyoung I remember that 'resource unavailable' error appears when inter/intra not setting on my machine. So it might be related to your issue, can you try inter=1 or 2 while intra=number_of_cores for configuration?

pandaoxin:
@waleedka It also benefits GPU, while quite a lot of TF Ops aren't supported by CUDA and only run on CPU. Besides, for MaskRCNN inference, CPU has high popularity and good performance. So in this case CPU can also benefits a lot.

pandaoxin:
@waleedka For the code you listed, I'm finding the method to set_session for keras outside model.py but failed, maybe a little confused? In fact, for more samples in addition to coco.py, I think it's more generic to set it in Config class.

pandaoxin:
@waleedka Any updates?

waleedka:
@pandaoxin Good point about the non-GPU operrations. But TensorFlow sets the default for these settings such that they're good for most users, right? I'm guessing you'd only need to change them if you have a unique setup. Does that make sense?

What was the issue you had with `set_session()`? I've used it in the past for other types of settings, and it worked for me when I call it right before creating the model.

dcyoung:
@pandaoxin Thanks. Unfortunately this did not solve my issue, but i may be having trouble with certain configurations of multiple gpus on my system in general, so can't say its related.

pandaoxin:
@waleedka Any updates?

yiqianglee:
@rgommers @waleedka This PR doesn't change thread configuration for Tensorflow, just expose the interface for user to tune the thread configuration (inter/intra), if you don't set these two configurations in script command, it won't impact anything, just useful for those user who want to tune performance with different inter/intra combination. Thanks!

PR #839, opened 2018-08-06
Fix bug: the para is not used for append_data method.

elias-1:
See https://github.com/matterport/Mask_RCNN/blob/master/mrcnn/utils.py#L342


waleedka:
It looks like `append_data()` is not used anywhere. I think it's an old function that was supposed to be removed but wasn't. Did you use it for anything? If not, I think a better fix is to remove the function.

waleedka:
I went ahead and remove `append_data()`. Closing this PR.

PR #873, opened 2018-08-20
Fix for UserWarning on anti-aliasing

cardboardcode:
The following modification made to **requirements.txt** removes the following user warning encountered during training. 
`UserWarning: Anti-aliasing will be enabled by default in skimage 0.15 to avoid aliasing artifacts when down-sampling images. warn("Anti-aliasing will be enabled by default in skimage 0.15 to " class_ids----------->`

**Rationale:** Minimizes compiler verbosity in terminal and makes epoch training indicators more obvious for effective debugging.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ scipy
 Pillow
 cython
 matplotlib
-scikit-image
+scikit-image==0.13.1
 tensorflow>=1.3.0
 keras>=2.0.8
 opencv-python


maxfrei750:
This issue has already been addressed in the PRs #674 and #756.

cardboardcode:
@maxfrei750 
Thank you for responding.
The UserWarning on anti-aliasing is still present in the current distribution, despite both PRs being closed.

maxfrei750:
@cardboardcode
I know, but I'm unsure, whether a downgrade of scikit-image is the best way to go, because it might be useful to utilize the anti_aliasing function to prevent artifacts. 

As soon as scikit-image 0.15 is released, anti_aliasing=True will be the default and the warning should no longer be thrown. However, for 0.14 it will stay and 0.14 is a LTS version and therefore might be used for quite a long time by people. Due to this reasons, the cleanest solution would be the one proposed by @waleedka in #674, i.e. implementation of a central resize function, which handles the different scikit-image versions.

cardboardcode:
@maxfrei750 
Noted. My modification is simply a hot fix, as opposed to the cleaner long-term fix by @waleedka. Would probably be best to wait for scikit-image 0.15 then. Current version of scikit-image is [0.14](http://scikit-image.org/docs/stable/)

Thank you for the clarification. This PR can be closed then.

maxfrei750:
@cardboardcode You're welcome. I think as a hotfix, your solution is very good, because it doesn't break the code for people using scikit-image<0.14 and I think scikit-image 0.13 should provide all the functionality we need for now. Therefore, it might make sense to merge the PR and remove the version requirement, as soon as someone found the time to write the central resize function.

waleedka:
I moved all `skimage.transform.resize()` calls to a central place, and added logic to use the right parameters depending on the version. This solves the problem without requiring downgrading skimage, and it makes it easier to fix other similar warnings if they come up in future versions. My update is here:
https://github.com/matterport/Mask_RCNN/commit/64020bd58f3772ebc97db920ce2725ce250a27c9 

Closing this PR now. Thank you for the help and discussion.

PR #877, opened 2018-08-21
Weight='last' not working (fixed) and creating unnecessary log directories while retraining (fixed)

ashuta03:
While trying to retrain the model using `weight='last'` argument, the existing code does not work as new log directory inside `logs` folder is created because of `self.set_log_dir()` in the `MaskRcnn` constructor. This creates a new log directory before checking for the `weight` argument which creates two problems:
1. For the `weight='last'` option because the directory returned by `model.find_last()` would be the newly created empty log directory instead of the actual last trained directory which  has all the required checkpoints. 
2. Instead of `weight='last'` if we give specific path to the checkpoint file like: `weight=path/to/existing/weight.h5` to start the training from the given checkpoint, an unnecessary new log directory is created, before continuing with the required checkpoint directory. 
Both problems are fixed by removing the `self.set_log_dir()` from the `MaskRcnn` constructor inside `model.py`.

### DIFF
--- a/mrcnn/model.py
+++ b/mrcnn/model.py
@@ -2283,7 +2283,6 @@ class MaskRCNN():
             self.config.NAME.lower()))
         self.checkpoint_path = self.checkpoint_path.replace(
             "*epoch*", "{epoch:04d}")
-        print('the log_dir and checkpoint path is ', self.log_dir, self.checkpoint_path)
 
     def train(self, train_dataset, val_dataset, learning_rate, epochs, layers,
               augmentation=None, custom_callbacks=None, no_augmentation_sources=None):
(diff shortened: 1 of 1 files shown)


maxfrei750:
This problem is also addressed by #823, which just moved the `self.set_log_dir()` to the training method. From my point of view that's the better solution.

ashuta03:
@maxfrei750 My bad! I saw the pull request only after i posted mine. It is a pretty effective solution. But in my defense, mine just requires a one line comment :D

maxfrei750:
@ashuta03 :smile:

waleedka:
Thanks @ashuta03. I merged https://github.com/matterport/Mask_RCNN/pull/823, which should solve the problem.

PR #878, opened 2018-08-21
Fixed comment on GPU_COUNT

maxfrei750:
Removed "For CPU training, use 1", because it doesn't seem to be correct.

### DIFF
--- a/mrcnn/config.py
+++ b/mrcnn/config.py
@@ -24,7 +24,7 @@ class Config(object):
     # experiment is running.
     NAME = None  # Override in sub-classes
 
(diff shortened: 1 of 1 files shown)


waleedka:
It's not clear to me how this fixes the comment. If someone uses  CPU only, then need to set GPU_COUNT = 1. Removing that comment makes it seem that when using a CPU then you set the count to 0.

maxfrei750:
@waleedka First of all: Nice to see you around this repo again!

I see what you mean, so I tried to fix the comment to include CPU usage.

PR #925, opened 2018-09-13
Fix the comment on output shape in RPN

keineahnung2345:


waleedka:
Good catch! Thank you.

Issue #928, opened 2018-09-13
 AttributeError: 'list' object has no attribute 'values'


eyildiz-ugoe:
As described in the balloon example, I've used VIA tool to annotate my images. I've used circles, polygons, etc. and gave each class its respective name.

![1](https://user-images.githubusercontent.com/35396058/45495438-8c6ee500-b773-11e8-9aa1-803026bf4db0.png)
![2](https://user-images.githubusercontent.com/35396058/45495439-8c6ee500-b773-11e8-9d1c-146e1bea9724.png)

Now that I want to load the dataset, again, using the same code, and I get the following error:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-9-048a8da58a4c> in <module>()
      1 # Load validation dataset
      2 dataset = components.ComponentsDataset()
----> 3 dataset.load_components(COMPONENTS_DIR, "val")
      4 
      5 # Must call before using the dataset

~/workspace/Mask_RCNN/samples/components/components.py in load_components(self, dataset_dir, subset)
    122             # the outline of each object instance. There are stored in the
    123             # shape_attributes (see json format above)
--> 124             polygons = [r['shape_attributes'] for r in a['regions'].values()]
    125 
    126             # load_mask() needs the image size to convert polygons to masks.

AttributeError: 'list' object has no attribute 'values'
```

What could be the problem? Has anyone tried loading the annotated data which contains circles? (not only polygons)

sajjad-taheri:
The samples only accept the polygons. That's why you get this error. You can either re-annotate the circles or use something like this to convert the circles to polygons in your json file:

```

import json
from pprint import pprint

with open('via_region_data(val).json') as f:
    data = json.load(f)

    for attr, val in data.items():
        for attr2, val2 in val.items():
            if attr2 == 'regions':
                for attr3, val3 in val2.items():
                    if val3['shape_attributes']['name'] == 'circle':
                        cx = val3['shape_attributes']['cx']
                        cy = val3['shape_attributes']['cy']
                        r = val3['shape_attributes']['r']
                        all_points_x = [cx, cx - 1.5 * r, cx, cx + 1.5 * r, cx]
                        all_points_y = [cy - 1.5 * r, cy, cy + 1.5 * r, cy, cy - 1.5 * r]
                        val3['shape_attributes']['cx'] = all_points_x
                        val3['shape_attributes']['cy'] = all_points_y

[...]

PR #943, opened 2018-09-18
Made compatible to new version of VIA JSON format

skt7:
VIA has changed JSON formatting in later versions. Now instead of a dictionary, "regions" has a list, see the issue #928


waleedka:
Thank you!

PR #940, opened 2018-09-17
add IMAGE_CHANNEL_COUNT class variable to config

moorage:
Adds IMAGE_CHANNEL_COUNT class variable to config to make it easier to use Mask_RCNN for non 3-channel images, like grayscale or RGB-D

PR #956, opened 2018-09-21
Determine IMAGE_CHANNEL_COUNT based on MEAN_PIXEL

maxfrei750:
Could be merged if #940 was merged.


waleedka:
Thanks, @maxfrei750. Chaning image depth is a significant change that few users need, but when they do, they need to make other significant changes along with it. So I prefer to have it be an explicit change that someone needs to make manually and be aware of its implications. So I'm inclined to skip this PR.

Issue #618, opened 2018-05-31
Discussion about the anchor center


HC-2016:
First, thank the authors for the well documented project.
Here, I have some confusion about the anchor as the followings:

According the codes in [https://github.com/matterport/Mask_RCNN/blob/master/mrcnn/utils.py#L607](url), the anchors in all feature maps start from (0, 0).
anchor centers in all feature maps:
scale1(stride=4): (0, 0), (0, 0), (0, 0), (0, 4), ... , (0, 8), .... , (0, 16), ....
scale2(stride=8): (0, 0), (0, 0), (0, 0), (0, 8), ... , (0, 16), ....
scale3(stride=16): (0, 0), (0, 0), (0, 0), (0, 16), ....
...
It's wonderful that some anchors in different scales are aligned.

However, in another project [https://github.com/smallcorgi/Faster-RCNN_TF](url) ,
the anchors in the single scale feature start from (7.5, 7.5) which depends on the size of the reference (0, 0, 15, 15) window. See the codes In the [https://github.com/smallcorgi/Faster-RCNN_TF/blob/master/lib/rpn_msr/proposal_layer_tf.py#L36](url)
anchor centers in the feature map(stride=16):
(7.5, 7.5), ...,  (7.5, 23.5), ... , (7.5, 39.5)

In my opinion, the second definition seems more reasonable. 
Taking the perceptual field into account, I think the anchor centers should start from (stride/2, stride/2).
 
I wonder whether the performance  would change with different anchor setting?

waleedka:
Good point! I agree that the center of the anchor would ideally be at the center of the receptive field. I don't know if it has a noticeable impact, though, because the difference is small relative to the size of the receptive field. The network probably learns to adjust for it.

If you get the chance to test it, I'd be very interested to learn if you notice any difference.

PR #957, opened 2018-09-21
adjust anchor center

keineahnung2345:
To the mentioned issue in https://github.com/matterport/Mask_RCNN/issues/618, this PR has adjusted the position of anchor centers.


waleedka:
Thanks! I haven't tested this yet, but I'm assuming it requires re-training the network (at least the RPN part), right?

Also, I'm curious, did you notice accuracy improvement with this change?

keineahnung2345:
@waleedka You are right, but I haven't tried it yet, too.

waleedka:
Okay, then I'll merge this a bit later with other updates that require retraining.

PR #961, opened 2018-09-22
fix the input shape of anchors in ProposalLayer

keineahnung2345:
fix the input shape of anchors in ProposalLayer and use "num_anchors" to replace "anchors" to avoid confusion

### DIFF
--- a/mrcnn/model.py
+++ b/mrcnn/model.py
@@ -259,9 +259,9 @@ class ProposalLayer(KE.Layer):
     box refinement deltas to anchors.
 
     Inputs:
-        rpn_probs: [batch, anchors, (bg prob, fg prob)]
-        rpn_bbox: [batch, anchors, (dy, dx, log(dh), log(dw))]
-        anchors: [batch, (y1, x1, y2, x2)] anchors in normalized coordinates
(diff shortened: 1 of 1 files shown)


mrcnn/model.py
```diff
@@ -259,9 +259,9 @@ class ProposalLayer(KE.Layer):
     box refinement deltas to anchors.
 
     Inputs:
-        rpn_probs: [batch, anchors, (bg prob, fg prob)]
-        rpn_bbox: [batch, anchors, (dy, dx, log(dh), log(dw))]
-        anchors: [batch, (y1, x1, y2, x2)] anchors in normalized coordinates
+        rpn_probs: [batch, num_anchors, (bg prob, fg prob)]
+        rpn_bbox: [batch, num_anchors, (dy, dx, log(dh), log(dw))]
+        anchors: [num_anchors, (y1, x1, y2, x2)] anchors in normalized coordinates
```
waleedka: I believe `anchors` has a batch dimension. The correct shape is [batch, anchors, (y1, x1, y2, x2)]

keineahnung2345: Thanks, I have corrected it.

PR #962, opened 2018-09-22
fix comment in DetectionTargetLayer

keineahnung2345:


keineahnung2345:
@waleedka I have changed it back to N.

PR #967, opened 2018-09-24
use smooth_l1_loss() to reduce code duplication

bill2239:
use smooth_l1_loss() to reduce code duplication in rpn_bbox_loss_graph


waleedka:
Thank you!

PR #970, opened 2018-09-25
Fix comment in FPN heads

keineahnung2345:
1. in PyramidROIAlign layer, use more specific pool_height, pool_width to replace height and width
2. fix the comment on output shape in fpn_classifier_graph
3. unify NUM_CLASSES and num_classes to NUM_CLASSES
4. unify boxes, num_boxes, num_rois, roi_count to num_rois
5. use more specific POOL_SIZE and MASK_ POOL_SIZE to replace pool_height and pool_width


waleedka:
Looks good. Thank you!

PR #1018, opened 2018-10-11
Fix #551

### LINKED ISSUE #551: AP not between 0 and 1 ?
So, i'm using `compute_ap` to compute a mean average precision in the following way (similar to the example in `train_shapes`) :

- APs = []
- For all images in my validation dataset
- - Load image
- - Detect on image
- - - If "rois" is empty, AP = 0
- - - Else  `AP, precisions, recalls, overlaps = utils.compute_ap()`
- - Append AP to APs
- mAP = mean(APs)

My problem is that a lot of AP values i compute are over 1.0 (example of an APs of a small set : `[2.0, 2.0, 2.0, 2.0, 2.0, 0.3333333333333333, 2.0, 0.05555555555555555, 1.375, 2.0, 2.0, 2.0, 0, 1.375, 2.0, 2.0, 2.0, 2.0, 1.5, 1.3, 2.0, 1.138888920346896, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 1.0, 2.0, 2.0, 1.5, 2.0, 2.0, 2.0, 1.25, 1.5, 2.0, 1.375, 1.6666666666666665, 2.0, 0.0, 0.25, 1.25, 2.0, 1.375, 1.375, 2.0, 2.0, 2.0, 0.5, 1.1666666666666667, 2.0, 2.0, 0.8333333333333333, 2.0, 1.5]` )

So my mAP is `mAP :  1.6073586750158326`, over 1.0 and i don't get it. Maybe it's because i have multiple detections for the same object? Maybe i don't understand what AP stands for and in this case i think i would like it very much if someone could help me :zipper_mouth_face: 

Thanks.

PS : i think I'm simply mistaking AP for precisions but now the same issue appears with my recalls, that are spread between 0 and 2 and it makes no sense to me.

marluca:
Valid index 0 is not regarded, leading to possibly more prediction_matches than grountruh_matches and recall values higher than 1.

### DIFF
--- a/mrcnn/utils.py
+++ b/mrcnn/utils.py
@@ -694,7 +694,7 @@ def compute_matches(gt_boxes, gt_class_ids, gt_masks,
         # 3. Find the match
         for j in sorted_ixs:
             # If ground truth box is already matched, go to next one
-            if gt_match[j] > 0:
+            if gt_match[j] > -1:
                 continue
             # If we reach IoU smaller than the threshold, end the loop
             iou = overlaps[i, j]


waleedka:
Good catch. Thank you!

PR #1023, opened 2018-10-14
Online Hard Example Mining (OHEM) for RPN

gaborvecsei:
- Online hard example mining implemented inside the `rpn_class_loss_graph`


gaborvecsei:
This is not the right implementation, and because of this I am closing the PR

ApoorvaSuresh:
Hi,
do you mean this has to be used for mrcnn_class_loss?

PR #1054, opened 2018-10-23
[Bug fix]compute_overlaps_masks cannot filter empty masks

keineahnung2345:
In compute_overlaps_masks(), the code checking the shape of masks1 and masks2 is not correct.
1. It should check masks's last dimension instead of first dimension to determine whether it's empty or not
2. When masks1 or masks2 is empty, it should return a zero array of shape (masks1.shape[-1], masks2.shape[-1])


waleedka:
You're right. Thank you. Merged.

PR #1058, opened 2018-10-23
simplify find_last()

keineahnung2345:
The weights of a model reside in model.log_dir, so the process of finding dir_name can be removed.
And there will be one more advantage: When there are multiple models running at same time, the result of find_last() will be unpredictable using the original version, and this is solved by the updated version.

### DIFF
--- a/mrcnn/model.py
+++ b/mrcnn/model.py
@@ -2069,27 +2069,15 @@ class MaskRCNN():
         Returns:
             The path of the last checkpoint file
         """
-        # Get directory names. Each directory corresponds to a model
-        dir_names = next(os.walk(self.model_dir))[1]
-        key = self.config.NAME.lower()
-        dir_names = filter(lambda f: f.startswith(key), dir_names)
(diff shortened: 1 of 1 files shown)


waleedka:
I think this breaks the `--weights=last` command line parameter.

When a `MaskRCNN` object is created, it sets the value of `log_dir` to a new directory that hasn't been created yet (in the expectation that you're about to run a new experiment). Then, if you call `load_weights()`, it updates `log_dir` to point to the directory of the weights you specified. 

Here is the catch: when you use `--weights=last`, you need to call `find_last()` to find the directory of the last experiment, then you call `load_weights()` and pass that directory to it, which then calls `set_log_dir()` to update `log_dir` to point to that directory. Updating `find_last()` to rely on `log_dir` creates a circular dependency that causes `log_dir` to always create a new directory rather than re-using the last one.

keineahnung2345:
@waleedka I got it.

PR #1061, opened 2018-10-24
use cv2 to draw rectangle in draw_box

keineahnung2345:
use cv2 to draw rectangle in draw_box
Note: cv2's coordinate system is different from that of numpy


keineahnung2345:
@waleedka I've noticed that `cv2` is not a required package in this project, should I reimplement this function using `skimage`?

PR #1065, opened 2018-10-25
Make run_graph() can be run in training mode

keineahnung2345:
The variable `inputs` is declared but not used in `run_graph()`.


keineahnung2345:
Update: Make run_graph() can be run in training mode
1. The argument become inputs and outputs
2. It can now run in training mode(useful in debugging), but only works when batch size equals to 1.

waleedka:
I found a place in inspect_nucleus_model.ipynb that uses the image_metas param. Which would break with this update. I wonder if we can change that notebook to pass the original image, and if so, then image_meta param is not needed anymore and can be removed to simplify. I'm hoping to get a bit of time to look into that in the next few days.

keineahnung2345:
@waleedka Yes, I noticed that point. I just want to make sure this PR will work and then start to apply changes to other notebooks. Maybe I should add a WIP flag.

keineahnung2345:
@waleedka There is only one notebook use the argument "image_metas", which is inspect_nucleus_model.ipynb. I have already made it compatible with the new run_graph.

PR #1067, opened 2018-10-25
Remove batch_size from data_generator's arguments

keineahnung2345:
The argument batch_size can be inferred from config, so it should be redundant.


waleedka:
There are a couple other places in Jupyter notebooks that call `data_generator()` to visualize data samples, and they pass in a `batch_size` that is usually different from the value in config. 

While this PR simplifies the code a bit by removing a function argument, I believe the flexibility we get in the notebooks justifies keeping it.

Issue #1044, opened 2018-10-20
 Predict problem on multi gpu (Input to reshape is a tensor with 24000 values, but the requested has 48000)


kawa23:
I try to  predict with  GPU_COUNT = 2, but got the issue:

<img width="1120" alt="screen shot 2018-10-20 at 8 38 15 pm" src="https://user-images.githubusercontent.com/39022504/47255773-9093ce00-d4a8-11e8-95a8-6a7f475f2a28.png">

`keras.__version__ = 2.1.5`  and  `tf.__version__=1.7.0`  and my config was:

```
config.BATCH_SIZE = 4
config.IMAGES_PER_GPU = 2
config.GPU_COUNT = 2 
```

And it worked fine with single gpu like that:

```
config.BATCH_SIZE = 2
config.IMAGES_PER_GPU = 2
config.GPU_COUNT = 1 
```

Tachyu:
same problem :(

kawa23:
> same problem :(

I  thought it cause by `DetectionLayer output reshape` , 

see here:
[https://github.com/matterport/Mask_RCNN/pull/1082](url)

hope this solve your problem

PR #1082, opened 2018-10-27
[Bug fix]: predict error on multi-gpu

kawa23:
Predict problem on multi gpu (Input to reshape is a tensor with 24000 values, but the requested has 48000) 

I thought  this issue, see also:( [#1044](https://github.com/matterport/Mask_RCNN/issues/1044) )caused by `DetectionLayer output reshape`

[https://github.com/matterport/Mask_RCNN/blob/master/mrcnn/model.py#L820](https://github.com/matterport/Mask_RCNN/blob/master/mrcnn/model.py#L820)
```
 return tf.reshape(
            detections_batch,
            [self.config.BATCH_SIZE, self.config.DETECTION_MAX_INSTANCES, 6])
```
was called before `parallel_model  merge` at the `detection func`
[https://github.com/matterport/Mask_RCNN/blob/master/mrcnn/model.py#L2043](https://github.com/matterport/Mask_RCNN/blob/master/mrcnn/model.py#L2043)

so I think, the `DetectionLayer output reshape`  was error, and it should be changed like that,
```
 return tf.reshape(
            detections_batch,
            [self.config.IMAGES_PER_GPU, self.config.DETECTION_MAX_INSTANCES, 6])
```
it worked for me on mulit-gpu(two)

### DIFF
--- a/mrcnn/model.py
+++ b/mrcnn/model.py
@@ -817,7 +817,7 @@ class DetectionLayer(KE.Layer):
         # normalized coordinates
         return tf.reshape(
             detections_batch,
-            [self.config.BATCH_SIZE, self.config.DETECTION_MAX_INSTANCES, 6])
+            [self.config.IMAGES_PER_GPU, self.config.DETECTION_MAX_INSTANCES, 6])
 
     def compute_output_shape(self, input_shape):
         return (None, self.config.DETECTION_MAX_INSTANCES, 6)


kawa23:
> @kawa23 Thanks for the pull request. I'm reviewing it and trying to trace the issue. A quick question about the error you reported in #1044: How come your tensor shape is 24000? I'd expect it to be 2400 since your batch size is 4.
> 
> ```
> BATCH_SIZE * DETECTION_MAX_INSTANCES * 6
> ```
> Did you happen to change `DETECTION_MAX_INSTANCES` from the default value of 100?
> 
> I acknowledge the error you got, but I'm suspecting that the issue might be somewhere else.

Yes, When I changed the `DETECTION_MAX_INSTANCES` from the default value of 100 to 2000,  I got this error:` Input to reshape is a tensor with 24000 values, but the requested has 48000
`, 

and the `DETECTION_MAX_INSTANCES ` was  100,  it came out ` Input to reshape is a tensor with 1200 values, but the requested has 2400`. 
I found that the `request reshape` is twice as` input reshape` based on several other data tests, the same as my `GPU_COUNT=2` setting.

And I also found that it was here [#L819](https://github.com/matterport/Mask_RCNN/blob/a0a2aa00f813ee61421d96b6f4c90e711b962a9d/mrcnn/model.py#L819) ~ [#L821](https://github.com/matterport/Mask_RCNN/blob/a0a2aa00f813ee61421d96b6f4c90e711b962a9d/mrcnn/model.py#L821)

 `return tf.reshape(
            detections_batch,
            [self.config.IMAGES_PER_GPU, self.config.DETECTION_MAX_INSTANCES, 6])`

throws the error when I tracking it.

May be I should test on  `GPU_COUNT=3` or more, but I just only have two.

Thanks for replying me, and I'll appreciate it if you find out and fix it.

lixindou2018:
@keineahnung2345 @kawa23 ，I tried to use multiple GPU to test Mask RCNN recently, but the speed did not improve. Could you post your code? thank you

tianyu-tristan:
> @kawa23 Thanks for the pull request. I'm reviewing it and trying to trace the issue. A quick question about the error you reported in #1044: How come your tensor shape is 24000? I'd expect it to be 2400 since your batch size is 4.
> 
> ```
> BATCH_SIZE * DETECTION_MAX_INSTANCES * 6
> ```
> Did you happen to change `DETECTION_MAX_INSTANCES` from the default value of 100?
> 
> I acknowledge the error you got, but I'm suspecting that the issue might be somewhere else.

Hi, thank you for ack on the error. I'm having the same issue (2 GPU, 2 images/GPU, no change of config so I'm 2400 as you stated here, instead of 24000). What's the proper fix to this?

vikiQiu:
> @kawa23 Thanks for the pull request. I'm reviewing it and trying to trace the issue. A quick question about the error you reported in #1044: How come your tensor shape is 24000? I'd expect it to be 2400 since your batch size is 4.
> 
> ```
> BATCH_SIZE * DETECTION_MAX_INSTANCES * 6
> ```
> 
> Did you happen to change `DETECTION_MAX_INSTANCES` from the default value of 100?
> 
> I acknowledge the error you got, but I'm suspecting that the issue might be somewhere else.

I get the similar problem as the up:
`Predict problem on multi gpu (Input to reshape is a tensor with 600 values, but the requested has 1200)`
I set DETECTION_MAX_INSTANCES as default 100, IMAGES_PER_GPU=2 and GPU_COUNT=1.
The problem is solved when I modify the code. It's quite helpful. I think it should be merge into master.

phongvu99 (approved):
Work like a charm!

Aashish-Gautam (commented):
It fixed my issue. Thanks for the help.
Not sure why this change is already not been done in model.py

Issue #726, opened 2018-06-27
Will well-trained RPN outputs only positive anchors during training?


jimmy15923:
First, thanks for well-documented and awesome implementation on Mask-RCNN!

I am working on nucleus data which have much more crowd nucleus compare to DSB 2018 data, but same goal to segment the nucleus part. One image contains lots of object, up to 1,500 masks on one image.

I have read model.py code for understanding how RPN works. For Proposal layer, it take anchors from RPN, sort it by foreground class score and apply NMS. So most of the ROIs will have high probability of  being foreground class.

```python
# Box Scores. Use the foreground class confidence. [Batch, num_rois, 1]
scores = inputs[0][:, :, 1]
ix = tf.nn.top_k(scores, pre_nms_limit, sorted=True, name="top_anchors").indices
scores = utils.batch_slice([scores, ix], lambda x, y: tf.gather(x, y), self.config.IMAGES_PER_GPU)
```

After Proposal layer, it will send those ROIs into DetectionTargetLayer and sample some positive ROIs and negative ROIs by comparing the IOU of Ground Truth. The sampled ROIs will be used to train the classifier and regressor. Mask-RCNN paper suggests that the ratio of positive ROI and negative ROI should be 1:3.

```python
# 1. Positive ROIs are those with >= 0.5 IoU with a GT box
positive_roi_bool = (roi_iou_max >= 0.5)
positive_indices = tf.where(positive_roi_bool)[:, 0]
# 2. Negative ROIs are those with < 0.5 with every GT box. Skip crowds.
negative_indices = tf.where(tf.logical_and(roi_iou_max < 0.5, no_crowd_bool))[:, 0]
```

My question is, what if our RPN is trained super well and it can catch almost every foreground class? if that happened, then the DetectionTargetLayer will get lots of positive ROIs and sample only few negative ROIs which might affect the training of classifier?

In my cases, many nucleus are pretty small and crowd, most of nucleus are side by side. So it's easy for RPN selecting to many positive anchors on IOU > 0.5, I am wondering this might hurt the result.

If I misunderstand anything please correct me and thanks again for awesome implementation!

waleedka:
[...]

PR #1088, opened 2018-10-29
Handle zero pad & Allow bg to train RPN

keineahnung2345:
Handle zero pad:

- fpn_classifier_graph: first use target_class_ids to find out the samples which are zero padding and finally zero out the result
- mrcnn_class_loss_graph: use target_class_ids to find out the samples which are zero padding and finally remove them from the loss
- add comments to detection_targets_graph, fpn_classifier_graph

Allow background images to train RPN
- build_rpn_targets: generate negative anchors for sample that doesn't have instances
- generate_random_rois: generate random rois for sample that doesn't have instances
- data_generator: allow images without any instances to train rpn(these kind of images will contribute to rpn_class_loss only)

### DIFF
--- a/mrcnn/model.py
+++ b/mrcnn/model.py
@@ -496,11 +496,11 @@ def detection_targets_graph(proposals, gt_class_ids, gt_boxes, gt_masks, config)
 
     Returns: Target ROIs and corresponding class IDs, bounding box shifts,
     and masks.
-    rois: [TRAIN_ROIS_PER_IMAGE, (y1, x1, y2, x2)] in normalized coordinates
+    rois: [TRAIN_ROIS_PER_IMAGE, (y1, x1, y2, x2)] in normalized coordinates. Zero padded.
     class_ids: [TRAIN_ROIS_PER_IMAGE]. Integer class IDs. Zero padded.
-    deltas: [TRAIN_ROIS_PER_IMAGE, (dy, dx, log(dh), log(dw))]
+    deltas: [TRAIN_ROIS_PER_IMAGE, (dy, dx, log(dh), log(dw))]. Zero padded.
     masks: [TRAIN_ROIS_PER_IMAGE, height, width]. Masks cropped to bbox
-           boundaries and resized to neural network output size.
+           boundaries and resized to neural network output size. Zero padded.
 
     Note: Returned arrays might be zero padded if not enough target ROIs.
     """
@@ -899,13 +899,15 @@ def build_rpn_model(anchor_stride, anchors_per_location, depth):
 ############################################################
 
 def fpn_classifier_graph(rois, feature_maps, image_meta,
-                         pool_size, num_classes, train_bn=True,
+                         pool_size, num_classes, 
+			 images_per_gpu,
+			 train_bn=True,
                          fc_layers_size=1024):
     """Builds the computation graph of the feature pyramid network classifier
     and regressor heads.
 
     rois: [batch, num_rois, (y1, x1, y2, x2)] Proposal boxes in normalized
-          coordinates.
+          coordinates. Zero padded.
     feature_maps: List of feature maps from different layers of the pyramid,
                   [P2, P3, P4, P5]. Each has a different resolution.
     image_meta: [batch, (meta data)] Image details. See compose_image_meta()
@@ -915,11 +917,18 @@ def fpn_classifier_graph(rois, feature_maps, image_meta,
     fc_layers_size: Size of the 2 FC layers
 
     Returns:
-        logits: [batch, num_rois, NUM_CLASSES] classifier logits (before softmax)
-        probs: [batch, num_rois, NUM_CLASSES] classifier probabilities
+        logits: [batch, num_rois, NUM_CLASSES] classifier logits (before softmax). Zero padded.
+        probs: [batch, num_rois, NUM_CLASSES] classifier probabilities. Zero padded.
         bbox_deltas: [batch, num_rois, NUM_CLASSES, (dy, dx, log(dh), log(dw))] Deltas to apply to
-                     proposal boxes
+                     proposal boxes. Zero padded.
     """
+    # first find which rois are padded
+    # non_zeros: [batch, num_rois]
+    _, non_zeros = utils.batch_slice(
+            rois,
+            lambda x: trim_zeros_graph(x, name="trim_rois"),
+            images_per_gpu)
+    
     # ROI Pooling
     # Shape: [batch, num_rois, POOL_SIZE, POOL_SIZE, channels]
     x = PyramidROIAlign([pool_size, pool_size],
@@ -951,6 +960,24 @@ def fpn_classifier_graph(rois, feature_maps, image_meta,
     s = K.int_shape(x)
     mrcnn_bbox = KL.Reshape((s[1], num_classes, 4), name="mrcnn_bbox")(x)
 
+    # finally zero out the results which are calculated from zero padded rois
+    # non_zeros: [batch, num_rois, num_classes]
+    non_zeros = K.tile(K.expand_dims(non_zeros, -1), (1,1,num_classes))
+
+    mrcnn_class_logits = KL.Lambda(lambda x: \
+                       K.tf.where(non_zeros, x, \
+                       K.zeros_like(x)),
+                       name="mrcnn_class_logits_trimed")(mrcnn_class_logits)
+    mrcnn_probs = KL.Lambda(lambda x: \
+                       K.tf.where(non_zeros, x, K.zeros_like(x)),
+                       name="mrcnn_class_trimed")(mrcnn_probs)
+    
+    # non_zeros: [batch, num_rois, num_classes, 4]
+    non_zeros = K.tile(K.expand_dims(non_zeros, -1), (1,1,1,4))
+    mrcnn_bbox = KL.Lambda(lambda x: \
+                       K.tf.where(non_zeros, x, K.zeros_like(x)),
+                       name="mrcnn_bbox_trimed")(mrcnn_bbox)
+    
     return mrcnn_class_logits, mrcnn_probs, mrcnn_bbox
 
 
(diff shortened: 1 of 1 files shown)


antimirov:
Wow, I totally forgot that Mask_RCNN skips images without instances. Even though I saw the balloon example where they delete those samples. I thought it was only in that example for some reason.

keineahnung2345:
Yes, I hope this PR could be reviewed ASAP. @waleedka

pmaksud:
With your code is that enough to include photos with an empty polygons list  and  an empty string class name in the balloon.py file (or equivalent)  so that the photo is taken into account in the loss, or do you need other modification to use BG picture. My model has struggle with false positive and I'm convince i could do way better if i could include full background picture. I don't know if i'm clear though.

keineahnung2345:
@pmaksud It should work fine with bg image input. BG images only contribute to rpn_class_loss and nothing else.

yghlc:
@keineahnung2345 Hi, thank you for the pull request. 
I am working on remote sensing images. Including the bg to train is important for the final results. I merge this pull request to my fork version. 
The training loss seemed good. However, when I performed inference used the trained model, the code predicted 100 instances on each image. Before I merging this pull request, each image only had a few or zero instances, which made sense. I think the number of "100" came from "max number of final detections" (was set as 100) in config.py. 
Could you check the code? Thank you!

Best,
Lingcao

pmaksud:
In my case, i included some bg images in my train and test set simultaneously and i tried to overfit them but i still get false postive which is very strange.

patrickcgray:
Is there any reason this hasn't been merged into the main branch? Seems like a useful addition, I'm assuming many domains would benefit from additional negative training samples.

cfengai:
I've tried this in hoping it could help learning from images without any masks (i.e. images only have "background"). Yet I had similar issues as others expressed in this thread. The new model with learning from BKGD images actually gave more false positive than before. This is exactly the opposite to what I want. The point of using BKGD images is to improve FP (maybe with a little cost of FN). This code doesn't seem to work very well for this purpose.

Seanspt:
> I've tried this in hoping it could help learning from images without any masks (i.e. images only have "background"). Yet I had similar issues as others expressed in this thread. The new model with learning from BKGD images actually gave more false positive than before. This is exactly the opposite to what I want. The point of using BKGD images is to improve FP (maybe with a little cost of FN). This code doesn't seem to work very well for this purpose.

Same for me.

adriaciurana:
I have had problems implementing these changes and I have not been able to train well. Can anyone confirm if he has achieved good results?
Is there an option to pass images with empty masks to simulate this behavior?
Thanks!

MathiasKahlen:
@adriaciurana did you ever find out? :) I am also struggling with false positives, and I think adding BG images could help aswell, since the model performs really well on positive samples.

adriaciurana:
> @adriaciurana did you ever find out? :) I am also struggling with false positives, and I think adding BG images could help aswell, since the model performs really well on positive samples.

Sorry for the delay, I apply some changes in the code in order to train the RPN with background images. 

In build_rpn_targets (line 1457) function of model.py add
`if gt_class_ids.shape[0] == 0:
        rpn_match = -1 * np.ones([anchors.shape[0]], dtype=np.int32)
        rpn_bbox = generate_random_rois(image_shape, \
            config.RPN_TRAIN_ANCHORS_PER_IMAGE, gt_class_ids, gt_boxes)
        return rpn_match, rpn_bbox`

In generate_random_rois (line 1571) function of model.py change:
`rois_per_box = int(0.9 * count / gt_boxes.shape[0])`
by
`rois_per_box = int(0.9 * count / (gt_boxes.shape[0] + 0.000001))`

In data_generation (line 1718) function of model.py remove:
`if not np.any(gt_class_ids > 0):
                continue`

I have not been able to perform all the appropriate tests, but we do not observe problems in learning.

On the other hand we want to do the training in two phases (we have not yet been able to try it):
- The whole set RPN + FPN.
- Only the FPN: Activate USE_RPN_ROIS = False (config.py) and random_rois = True, detection_targets = True (model.py)
That way the FPN receives negatives and has the ability to classify better. False positives are given by an RPN error and because the FPN does not learn negatives well.

MathiasKahlen:
Thanks a lot, I'll have to try that out when I get the time for it :) I will get back with updates on how it goes.

luxedo:
Thanks!

K-M-Ibrahim-Khalilullah:
> 
> 
> Since MaskRCNN is composed of many different parts, so it's hard to say which part of the entire model goes wrong. There is another issue discussing the interaction of RPN, ProposalLayer and DetectionTargetLayer: #726, but we need more evidences. One can use the revised run_graph: #1065 to check output of each stage in **training mode**, so more detailed analysis will be possible.

@keineahnung2345 

Thanks for you comment. would you share code segment please to use run_graph? I want to check output of each stage during training?

konstantin-frolov:
> > @adriaciurana did you ever find out? :) I am also struggling with false positives, and I think adding BG images could help aswell, since the model performs really well on positive samples.
> 
> Sorry for the delay, I apply some changes in the code in order to train the RPN with background images.
> 
> In build_rpn_targets (line 1457) function of model.py add
> `if gt_class_ids.shape[0] == 0: rpn_match = -1 * np.ones([anchors.shape[0]], dtype=np.int32) rpn_bbox = generate_random_rois(image_shape, \ config.RPN_TRAIN_ANCHORS_PER_IMAGE, gt_class_ids, gt_boxes) return rpn_match, rpn_bbox`
> 
> In generate_random_rois (line 1571) function of model.py change:
> `rois_per_box = int(0.9 * count / gt_boxes.shape[0])`
> by
> `rois_per_box = int(0.9 * count / (gt_boxes.shape[0] + 0.000001))`
> 
> In data_generation (line 1718) function of model.py remove:
> `if not np.any(gt_class_ids > 0): continue`
> 
> I have not been able to perform all the appropriate tests, but we do not observe problems in learning.
> 
> On the other hand we want to do the training in two phases (we have not yet been able to try it):
> 
> * The whole set RPN + FPN.
> * Only the FPN: Activate USE_RPN_ROIS = False (config.py) and random_rois = True, detection_targets = True (model.py)
>   That way the FPN receives negatives and has the ability to classify better. False positives are given by an RPN error and because the FPN does not learn negatives well.

It's work fine. Thank you!!!

Spritaro:
I tried to use this PR for a single-class instance segmentation task, but the trained model produced more false positives. 

I looked up mrcnn_class_loss_graph and noticed that ROIs are removed from loss calculation if corresponding target_class_ids is 0. I know it removes zero-padding, but isn't BG class id 0 too?

I'm not sure if this is applicable to other tasks, but I was able to reduce false positive by changing the following line 
```non_zeros = tf.cast(tf.greater(target_class_ids, 0), 'float32')``` 
to
```non_zeros = tf.cast(tf.math.not_equal(tf.reduce_sum(pred_class_logits, axis=-1), 0), 'float32')```

konstantin-frolov:
> `non_zeros = tf.cast(tf.greater(target_class_ids, 0), 'float32')`

Hi, i can't find that string in model.py. Where you find it?

Spritaro:
@konstantin-frolov 
It's from files changed in this pull request. You can also find it here https://github.com/keineahnung2345/Mask_RCNN/blob/168ca5cdb7f5656722ec6ed2dd67451cde994421/mrcnn/model.py#L1118

konstantin-frolov:
> It's from files changed in this pull request. You can also find it here 

Thank's. I used approach from @adriaciurana for multi class detection. But when try detect single class (BG and object) only BG images have no effect for FP.

bertsky:
> I tried to use this PR for a single-class instance segmentation task, but the trained model produced more false positives.
> 
> I looked up mrcnn_class_loss_graph and noticed that ROIs are removed from loss calculation if corresponding target_class_ids is 0. I know it removes zero-padding, but isn't BG class id 0 too?
> 
> I'm not sure if this is applicable to other tasks, but I was able to reduce false positive by changing the following line
> `non_zeros = tf.cast(tf.greater(target_class_ids, 0), 'float32')`
> to
> `non_zeros = tf.cast(tf.math.not_equal(tf.reduce_sum(pred_class_logits, axis=-1), 0), 'float32')`

@Spritaro that is precisely the problem with this PR IMO. By filtering zero-valued `target_class_ids` out of the `mrcnn_class_loss_graph`, the model cannot learn to predict BG for bad proposals anymore – which becomes critical after allowing empty images in the training loop, or if you have images with few/sparse instances.

Your fix works because the class logits for true (BG) zeros are never zero _everywhere_. I have dealt with this myself by looking at the (indices of) all-zero bbox predictions. (Also, the same thing needs to be done for `mrcnn_bbox_loss_graph`.)

But there's one further ingredient required to learn BG well and avoid FP in my experience: In `detection_targets_graph` when positive targets are filled up with negative targets up to `ROI_POSITIVE_RATIO`, the problem is that if there are _too few_ positives (again from no or too few GT instances) then that ratio won't hold and there will be too few or no negatives. So the targets will simply be padded with zeros. The only way to fix this is to ensure a constant number of negatives.

You can find an implementation of that [here](https://github.com/bertsky/Mask_RCNN/commit/168ca5cdb7f5656722ec6ed2dd67451cde994421).

The [recipe](https://github.com/matterport/Mask_RCNN/pull/1088#issuecomment-541563405) proposed by @adriaciurana of running with `random_rois = True` at a later stage to make the FPN robust to bad proposals is still valid (if not as pressing anymore) though, especially if the RPN gets "too good" early on.

philippmaa:
Hi everybody,
first of all thanks for all the insightful information and implementations!
Currently, I am training a model to detect objects of only one class (If present on the image). I came across the same issue of having a model performing quite well when trained on images with at least one instance of the object present. However, it predicts a rather large amount of false positives, i.e. detects objects where no objects are.
I have tried the suggestions of this PR.  Unfortunately, after training now my model predicts even more false positives and all the predictions have detection confidence of 1.0.
Did anybody else encounter such behavior?

MatchaCookies:
> 
@bertsky hi thank you very much for this. When training with your code, do I need to put background images or images with undesired objects in my train and val dataset? I am training with 1 class only: which is concrete crack. My trained model apparently detects other objects such as humans, hands, even straight line or shadows as "crack". So may false positives. 

Also, do I need to change anything else aside from the model.py? 
I'm really new to cnn and such.

Issue #934, opened 2018-09-16
Visualize Activations


amin3511:
hi everyone 
i run inspect_balloon_model.ipynb but in the Visualize Activations section, i get this error:
InvalidArgumentError                      Traceback (most recent call last)
<ipython-input-36-8bd731e4a663> in <module>()
      6     ("res4w_out",          model.keras_model.get_layer("res4w_out").output),  # for resnet100
      7     ("rpn_bbox",           model.keras_model.get_layer("rpn_bbox").output),
----> 8     ("roi",                model.keras_model.get_layer("ROI").output),
      9 ])

~/Downloads/Mask_RCNN/mrcnn/model.py in run_graph(self, images, outputs, image_metas)
   2687         if model.uses_learning_phase and not isinstance(K.learning_phase(), int):
   2688             model_in.append(0.)
-> 2689         outputs_np = kf(model_in)
   2690 
   2691         # Pack the generated Numpy arrays into a a dict and log the results.

~/anaconda3/envs/MaskRCNN/lib/python3.6/site-packages/keras/backend/tensorflow_backend.py in __call__(self, inputs)
   2659                 return self._legacy_call(inputs)
   2660 
-> 2661             return self._call(inputs)
   2662         else:
   2663             if py_any(is_tensor(x) for x in inputs):

~/anaconda3/envs/MaskRCNN/lib/python3.6/site-packages/keras/backend/tensorflow_backend.py in _call(self, inputs)
   2628                                 feed_symbols,
   2629                                 symbol_vals,
-> 2630                                 session)
   2631         fetched = self._callable_fn(*array_vals)
   2632         return fetched[:len(self.outputs)]

~/anaconda3/envs/MaskRCNN/lib/python3.6/site-packages/keras/backend/tensorflow_backend.py in _make_callable(self, feed_arrays, feed_symbols, symbol_vals, session)
   2580         callable_opts.target.append(self.updates_op.name)
   2581         # Create callable.
-> 2582         callable_fn = session._make_callable_from_options(callable_opts)
   2583         # Cache parameters corresponding to the generated callable, so that
   2584         # we can detect future mismatches and refresh the callable.

~/anaconda3/envs/MaskRCNN/lib/python3.6/site-packages/tensorflow/python/client/session.py in _make_callable_from_options(self, callable_options)
   1478     """
   1479     self._extend_graph()
-> 1480     return BaseSession._Callable(self, callable_options)
   1481 
   1482 

[...]

PR #1211, opened 2019-01-09
Fix visualize activations

keineahnung2345:
```python
# Get activations of a few sample layers
activations = model.run_graph([image], [
    ("input_image",        model.keras_model.get_layer("input_image").output)
])
```
leads to the error:
> InvalidArgumentError: input_image:0 is both fed and fetched.

Revise the code according to https://stackoverflow.com/questions/39307108/placeholder-20-is-both-fed-and-fetched.

This PR fixes https://github.com/matterport/Mask_RCNN/issues/934.


waleedka:
Good fix. Thank you!

Issue #331, opened 2018-03-14
[Issue] Incompatible shapes during training of own dataset


Gloupys:
Hello again,

I successfully managed to use my dataset but the problem apparently is with the shapes inbetween layers... 

I get the following error:

InvalidArgumentError (see above for traceback): Incompatible shapes: [8,28,28] vs. [8,512,512]
	 [[Node: mrcnn_mask_loss/logistic_loss/mul = Mul[T=DT_FLOAT, _device="/job:localhost/replica:0/task:0/device:CPU:0"](mrcnn_mask_loss/Log, mrcnn_mask_loss/Gather_1)]]

I did change the batch size to see if there were any changes, and the shapes switch to [8,28,28] to [6,28,28] when I change the batch size from 16 to 8. 

Here are screens of the configuration and the error (partly):

![meh2](https://user-images.githubusercontent.com/12952763/37407595-58574a5a-279a-11e8-9365-64dc40d2b3fc.PNG)
![argh](https://user-images.githubusercontent.com/12952763/37407596-588d375a-279a-11e8-964d-413803a0bc6d.PNG)

I resized the images to be (512,512) as well. I have no clue on how to fix that...
Thanks for your help.

Mirodil:
I have the same issue but mine related to `MASK_SHAPE`

pesekon2:
I am facing the same error message:
`InvalidArgumentError (see above for traceback): Incompatible shapes: [256,4] vs. [68,4]`

The interesting thing is that I am facing it only when using `GPU_COUNT>2`. When using just one or two GPUs it runs without any problem. 

PS: My images are of size 768*768

keineahnung2345:
You need to add more deconv layers in build_fpn_mask_graph() to get bigger mask.
Please check: https://github.com/matterport/Mask_RCNN/issues/635.

rsumit123:
Did anyone find th solution for this

PR #1219, opened 2019-01-11
Configurable mask branch

keineahnung2345:
This PR makes MASK_SHAPE configurable, and the structure of mask branch will be adjusted accordingly. 
It also adds assertion for easier debugging, ref: https://github.com/matterport/Mask_RCNN/issues/331.

One can now set MASK_SHAPE as [28,28], [56,56], [112, 112], [224, 224], etc, given MASK_POOL_SIZE is 14.

### DIFF
--- a/mrcnn/model.py
+++ b/mrcnn/model.py
@@ -955,7 +955,7 @@ def fpn_classifier_graph(rois, feature_maps, image_meta,
 
 
 def build_fpn_mask_graph(rois, feature_maps, image_meta,
-                         pool_size, num_classes, train_bn=True):
+                         pool_size, num_classes, num_deconv_layers, train_bn=True):
     """Builds the computation graph of the mask head of Feature Pyramid Network.
 
     rois: [batch, num_rois, (y1, x1, y2, x2)] Proposal boxes in normalized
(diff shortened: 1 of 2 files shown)


mrcnn/model.py
```diff
@@ -999,8 +1000,10 @@ def build_fpn_mask_graph(rois, feature_maps, image_meta,
                            name='mrcnn_mask_bn4')(x, training=train_bn)
     x = KL.Activation('relu')(x)
 
-    x = KL.TimeDistributed(KL.Conv2DTranspose(256, (2, 2), strides=2, activation="relu"),
-                           name="mrcnn_mask_deconv")(x)
+    for i in range(num_deconv_layers):
+        x = KL.TimeDistributed(KL.Conv2DTranspose(256, (2, 2), strides=2, activation="relu"),
+                           name="mrcnn_mask_deconv" + str(i+1))(x)
```
waleedka: The first transposed conv should have the same name as before. In other words, `mrcnn_mask_deconv` rather than `mrcnn_mask_deconv1`. Otherwise, it will prevent loading the pre-trained COCO weights.

waleedka: The layer name should stay the same for the first layer. In other words, `mrcnn_mask_deconv` rather than `mrcnn_mask_deconv1`. This allows loading the pre-trained COCO weights when `num_deconv_layers` is 1.

mrcnn/model.py
```diff
@@ -999,8 +1000,10 @@ def build_fpn_mask_graph(rois, feature_maps, image_meta,
                            name='mrcnn_mask_bn4')(x, training=train_bn)
     x = KL.Activation('relu')(x)
 
-    x = KL.TimeDistributed(KL.Conv2DTranspose(256, (2, 2), strides=2, activation="relu"),
-                           name="mrcnn_mask_deconv")(x)
+    for i in range(num_deconv_layers):
```
waleedka: How about adding a comment explaining that the official paper only has one transposed conv, and that additional layers is an extra feature not described in the paper?

mrcnn/model.py
keineahnung2345: I've change the first deconv layer's name to "mrcnn_mask_deconv".

keineahnung2345: I've change the first deconv layer's name to "mrcnn_mask_deconv".

mrcnn/model.py
keineahnung2345: comment added right after the explanation of `num_deconv_layers`.

PR #1227, opened 2019-01-14
remove unused code

yanfengliu:
## This PR removes the following unused code:

in `mrcnn/model.py`:
(line 529) `crowd_masks = tf.gather(gt_masks, crowd_ix, axis=2)`

in `mrcnn/visualize.py`:
(line 143 & 439) `x = random.randint(x1, (x1 + x2) // 2)`

## This PR modifies the following redundant code:
in `mrcnn/model.py`:
(line 2264) `self.epoch = int(m.group(6)) - 1 + 1` **changes to** `self.epoch = int(m.group(6))`

### DIFF
--- a/mrcnn/model.py
+++ b/mrcnn/model.py
@@ -526,7 +526,6 @@ def detection_targets_graph(proposals, gt_class_ids, gt_boxes, gt_masks, config)
     crowd_ix = tf.where(gt_class_ids < 0)[:, 0]
     non_crowd_ix = tf.where(gt_class_ids > 0)[:, 0]
     crowd_boxes = tf.gather(gt_boxes, crowd_ix)
-    crowd_masks = tf.gather(gt_masks, crowd_ix, axis=2)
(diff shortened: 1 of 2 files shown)


mrcnn/model.py
```diff
@@ -2261,7 +2260,7 @@ def set_log_dir(self, model_path=None):
                                         int(m.group(4)), int(m.group(5)))
                 # Epoch number in file is 1-based, and in Keras code it's 0-based.
                 # So, adjust for that then increment by one to start from the next epoch
-                self.epoch = int(m.group(6)) - 1 + 1
```
waleedka: The `-1 + 1` is intentional to clarify the comment above it. To make it clear that we're subtracting 1 then adding 1.

waleedka:
The PR looks good. Thanks! The only issue is the comment I left above. I can see how it could be confusing, but I also think it could be more confusing without the explicit -1 + 1.

waleedka:
Thank you!

PR #1266, opened 2019-01-31
Update README.md, dependencies installation should happen after repos…

pbxqdown:
Update README.md, dependencies installation should happen after repository clone.


waleedka:
Thanks. I believe this was already done in PR https://github.com/matterport/Mask_RCNN/pull/1267 .

PR #1267, opened 2019-02-01
Reversed order (1) and (2) in installation; Fixes #1255

### LINKED ISSUE #1255: Installation instructions 1. and 2. needs swap in readme.md
I think it's pretty obvious to people but still deserves attention. Since,
1. should be. "Clone repository". Then
2. should be "Install requirements". 

image

manasbedmutha98:
Minor  fix. Reference to #1255, order has been made to clone first then use `pip install`

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,11 +179,11 @@ If you use Docker, the code has been verified to work on
 
 
 ## Installation
-1. Install dependencies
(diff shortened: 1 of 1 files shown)


waleedka:
Thank you!

PR #1268, opened 2019-02-04
add warning to load_mask of dataset class

xelmirage:
as described in https://github.com/matterport/Mask_RCNN/pull/1207/files


waleedka:
Thank you! This is merged, and I'll close the other one.

PR #1286, opened 2019-02-14
Multiple Anchors have same IOU

karanchahal:
Fixed the Todo for matching multiple anchors with the same IOU, and setting their value to 1. The previous version only picked the first anchor that had a maximum IOU. Now, if multiple anchors had the same IOU < 0.7, all of them will be set to a foreground class.


waleedka:
Thank you. Simple and clever solution.

karanchahal:
It's an honour !

PR #1305, opened 2019-02-25
Fix typos in inspect_data notebook

crystaldust:
- Change 'feature meap' into 'feature map'(looks like a typo)
- Replace "256\*256\*3" with "256x256x3", since the '*' won't be displayed in Markdown


waleedka:
Thank you!

Issue #1422, opened 2019-04-09
Bug in compute_matches


christianlandgraf:
Hi,

I have a question/problem regarding the compute_matches Function in utils.py (line 656)
Why do we sort the predictions from high to low?

```
indices = np.argsort(pred_scores)[::-1]
pred_boxes = pred_boxes[indices]
pred_class_ids = pred_class_ids[indices]
pred_scores = pred_scores[indices]
pred_masks = pred_masks[..., indices]

# Compute IoU overlaps [pred_masks, gt_masks]
overlaps = compute_overlaps_masks(pred_masks, gt_masks)
```

This is already done in the DetectionLayer. In most cases, this is unnecessary but does not lead to an error.

But if there are two or more detections with the same score,  ```np.argsort(pred_scores)[::-1] ``` gives a different sorting.

For example, if ```x = [0.99,0.99,0.99]```, then: ```np.argsort(x)[::-1] = [2,1,0]```

The order of ```pred_masks``` is changed and the output ```pred_match``` computes matches for the **changed array**(!), and not for the original array. Since we do not return the changed error, the matching is wrong

I suppose, this will affect the AP calculation as well.

PR #1436, opened 2019-04-14
Fix Bug in compute_matches

christianlandgraf:
Fix issue #1422. Removed the sorting in ```compute_matches(..)```

Predictions are already sorted in the DetectionLayer. 

If there are two or more detections with the same score,  ```np.argsort(pred_scores)[::-1] ``` gives a different sorting.

For example, if ```x = [0.99,0.99,0.99]```, then: ```np.argsort(x)[::-1] = [2,1,0]```

The order of ```pred_masks``` is changed and the output ```pred_match``` computes matches for the **changed array**(!), and not for the original array. We do not return the changed ```pred_masks```. Therefore, the returned matching is wrong.


SalahAdDin:
:+1: 
Please merge it.

PR #1444, opened 2019-04-17
fix minor typo in balloon mask_rcnn example

vinitra-zz:
As title. Change "ballon" into "balloon"


vinitra-zz:
@waleedka, quick ping for review!

Issue #715, opened 2018-06-24
mrcnn expects wrong shape for input image


rbavery:
Hello! I'm trying to train mrcnn on 8 channel imagery to detect agriculture. (channels are RGBN for on/off growing season). Here a notebook with my code: https://github.com/rbavery/deepimagery/blob/master/notebooks/Train_WV2_MRCNN.ipynb

I've followed the steps on the wiki for adjusting the mrcnn config and dataset classes for more than 3 channels, except for the last step since I don't want to use pregenerated weights and want the model to initialize those randomly: https://github.com/matterport/Mask_RCNN/wiki

The strange error is below, somewhere, something is expecting an array of shape (None,None, 3) but gets shape (256, 256, 8) instead which is the shape it should be expecting. You'll see that I've set the IMAGE_SHAPE channel to accept N, changed MEAN_PIXEL for 8 channels, and changed load_image to accept 8 channels.

I've tried to dig through the stacktrace to find why it would expect (None,None, 3) but am quite lost. Any help/suggestions are appreciated. I'm also confused why it is throwing `RuntimeError: sequence argument must have length equal to input rank` and if this issue is seperate or not from the other issue.

Please let me know if you need anymore information.

```
Train all layers
WARNING:tensorflow:From /opt/anaconda3/envs/maskrcnn/lib/python3.6/site-packages/tensorflow/contrib/learn/python/learn/datasets/base.py:198: retry (from tensorflow.contrib.learn.python.learn.datasets.base) is deprecated and will be removed in a future version.
Instructions for updating:
Use the retry module or similar alternatives.

Starting at epoch 0. LR=0.001

Checkpoint Path: /home/rave/tana-crunch/waves/deepimagery/data/raw/wv2/models/wv2-gridded20180624T1338/mask_rcnn_wv2-gridded_{epoch:04d}.h5
Selecting layers to train
conv1                  (Conv2D)
bn_conv1               (BatchNorm)
res2a_branch2a         (Conv2D)
bn2a_branch2a          (BatchNorm)
res2a_branch2b         (Conv2D)
bn2a_branch2b          (BatchNorm)
res2a_branch2c         (Conv2D)
res2a_branch1          (Conv2D)
bn2a_branch2c          (BatchNorm)
bn2a_branch1           (BatchNorm)
res2b_branch2a         (Conv2D)
bn2b_branch2a          (BatchNorm)
res2b_branch2b         (Conv2D)
bn2b_branch2b          (BatchNorm)
[...]

PR #1481, opened 2019-05-02
Fix for mask rcnn nucleus reading mask as 3 channel image

lijo381:
https://github.com/matterport/Mask_RCNN/issues/715#issuecomment-487954082

### DIFF
--- a/mrcnn/model.py
+++ b/mrcnn/model.py
@@ -1687,14 +1687,15 @@ def data_generator(dataset, config, shuffle=True, augment=False, augmentation=No
 
     # Keras requires a generator to run indefinitely.
     while True:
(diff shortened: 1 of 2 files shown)


NerdToMars (approved):
clear

andreasalamanos:
Could you please provide a description of what was wrong, and what did you change?
It means that I have to download and install again in order to use the update?

Thanks

lijo381:
> Could you please provide a description of what was wrong, and what did you change?
> It means that I have to download and install again in order to use the update?
> 
> Thanks
please look into this:
https://github.com/matterport/Mask_RCNN/issues/715#issuecomment-487954082

PR #1506, opened 2019-05-16
add class weight support for mrcnn heads

spotiris:
I have added class weight supports for the MRCNN heads through the member function MRCNN.set_class_weights(class_weights).
Note that this does not affect the RPN at this time.

I have successfully tested this on the shapes dataset, but nothing extensive.

### DIFF
--- a/mrcnn/model.py
+++ b/mrcnn/model.py
@@ -1009,11 +1009,11 @@ def build_fpn_mask_graph(rois, feature_maps, image_meta,
 #  Loss Functions
 ############################################################
 
-def smooth_l1_loss(y_true, y_pred):
+def smooth_l1_loss(y_true, y_pred, weights=1.0):
     """Implements Smooth-L1 loss.
     y_true and y_pred are typically: [N, 4], but could be any shape.
     """
-    diff = K.abs(y_true - y_pred)
+    diff = K.abs(y_true - y_pred) * weights
     less_than_one = K.cast(K.less(diff, 1.0), "float32")
     loss = (less_than_one * 0.5 * diff**2) + (1 - less_than_one) * (diff - 0.5)
     return loss
@@ -1074,7 +1074,7 @@ def rpn_bbox_loss_graph(config, target_bbox, rpn_match, rpn_bbox):
 
 
 def mrcnn_class_loss_graph(target_class_ids, pred_class_logits,
-                           active_class_ids):
+                           active_class_ids, class_weights):
     """Loss for the classifier head of Mask RCNN.
 
     target_class_ids: [batch, num_rois]. Integer class IDs. Uses zero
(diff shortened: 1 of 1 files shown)


andreasalamanos:
Hi,
could you please explain a bit more what you have done and most importantly why?
Thanks

spotiris:
> Hi,
> could you please explain a bit more what you have done and most importantly why?
> Thanks

I have added a variable class_weights of size [num_classes], which scales the losses of the mrcnn heads in the following way:
* mrcnn_bbox loss: in the smooth L1 loss error term
* mrcnn_class_loss: through tf.nn.sparse_softmax_cross_entropy, also added a switch case when no IDs are predicted, to avoid NaNs.
* mrcnn_mask_loss: by multiplying the result of K.binary_cross_entropy by the weights

The reason I have introduced this is to control the network to pay more and/or less attention to particular classes, as you would with large class imbalances, or in my case in the research topic of lifelong machine learning.

Marssouss:
Hi, I have somes questions about your modifications.
I don't understand if class_weights is calculated automaticaly with config.NUM_CLASSES or if I need to create an array who contain weights for each classes.
Thanks for your reply.

spotiris:
> Hi, I have somes questions about your modifications.
> I don't understand if class_weights is calculated automaticaly with config.NUM_CLASSES or if I need to create an array who contain weights for each classes.
> Thanks for your reply.

The class weights cannot be calculated automatically as the purpose and methods to do so are various.
You need to call the `set_class_weights` method of a MRCNN instance, with the `weights` being a numpy array of shape `(NUM_CLASSES,)`. Each element is the associated class weight for that class ID.
Take note that class 0 is always background, and must be non zero.

Marssouss:
Ah okay ! :)
Thanks for you reply ;)

HAMZARaouia:
First, thanks for the work you proposed to solve the class imbalance problem.

In fact, I applied the modifications you proposed, the maskrcnn model has learnt nothing as the loss fct looks the following: 

image

I went back to your code and tried to understand every modification you made.

I have a question about this line of code in **model.py** (one of the modifcation you proposed): 

image

Can you explain why you applied the weights to the **input_image** ?

The class weights are meant to interfere only in the loss fct at the end of the network, why applying them at the beginning of your network. 

Thank you

spotiris:
Hi @HAMZARaouia,

I have continued work on this and found one issue, however it was not related to your question.

Firstly the reason that the class weights _appear_ to be applied to the input image is the same reason that the repo author does the same [here](https://github.com/matterport/Mask_RCNN/blob/master/mrcnn/model.py#L1934).
It is to allow a tf.Variable to work in the Keras architecture, it does not actually have any influence on the input image.

However I have by accident divided the class weights by the wrong denominator [here](https://github.com/matterport/Mask_RCNN/pull/1506/files#diff-312c7e001d14bbb7ce5f8978f7b04cc3R1116).
I can confirm that this was in essence making class contribute very little to the optimization.

I will push the updated code fixing it shortly.
Please let me know if this works for your application!
I have had successful results with mine.

Edit: I've tested it for errors and the results are below:
image
The blue, red and green lines at the top of the loss graph are the erroneous implementation.
The gray and orange lines are the origin/master copy (no class weights) and the cyan line is the updated code I'm pushing today, with the class weights set to 1.
Unfortunately I do not have time right now to check it with imbalanced data and class weights enabled.

Apologies for being lax in pushing the code.

HAMZARaouia:
Hi @spotiris , 

Thanks for your answer and for the information you gave. The results shown on the graph look very promising.

So in order to make it work correctly, you had changed the denominator [here](https://github.com/matterport/Mask_RCNN/pull/1506/files#diff-312c7e001d14bbb7ce5f8978f7b04cc3R1116).

What is then the correct denominator you applied ? 

Thank you in advance for your help !

spotiris:
After some more testing I've found what I think to be a problem in the mrcnn_mask_loss, I have disabled the class balance for that loss.
NaNs frequently appear after around 3 epochs of training heads.

Results are below for the dummy dataset "shapes" consisting of 1000 training images and 100 validation images.
In all experiments below the same class balanced validation set is used, the training set is varied between a balanced and imbalanced set.
Although there are only three object classes, the first class weight represents the background class.
The class weights are varied to equal weighting (1/1/1/1), weighting according to the objects (0.099/0.099/0.985/0.099), and according to the objects but with the background class set to 1 (1/0.099/0.985/0.099).
It looks like the class weight, even without its incorporation into mrcnn_mask_loss is beneficial.


|  Code Branch  | Train Set Balance |      Class Weights      |  Color  |
|:-------------:|:-----------------:|:-----------------------:|:-------:|
| origin/master |   even |           N/A           |   cyan  |
| origin/master |  0.425/0.05/0.425 |           N/A           |  orange |
|      this     |   even  |         1/1/1/1         |   red   |
|      this     |  0.425/0.05/0.425 | 0.099/0.099/0.985/0.099 | magenta |
|      this     |  0.425/0.05/0.425 |   1/0.099/0.985/0.099   |  green  |

image

I'll clean and post the code today.

hoangphucITJP:
Hi @spotiris,
I see that the loss of your modified model is lower than the original. That means the performance is worse right? What can be the reason for that degradation?

PR #1507, opened 2019-05-16
Return history object after the training.

maxfrei750:
Gives users access to the training history.


spotiris:
I've done this also, it is very useful.

One thing to note is that `history` has a `history.model` member, which points to the model.
This leads to python garbage collection not releasing the GPU handles for the model, if you delete `model` without deleting `history`.

maxfrei750:
@spotiris That's very good to know. Thank's for the advice!

PR #1559, opened 2019-06-13
New Use Case: Drivable area detector.

kmh4321:
Added a new example/use case under `samples/` for detecting driveable area.


kmh4321:
cc: @waleedka

PR #1608, opened 2019-07-03
MASK RCNN on clevr Dataset

rimamittal:
Clevr Dataset (https://cs.stanford.edu/people/jcjohns/clevr/). Used Mask RCNN to generate object proposals.


Cuberick-Orion:
Hi, would it be possible for you to share the pretrained model on clevr as well? Many thanks!

Cuberick-Orion:
Hi rimamittal,

Thanks for your quick response to my comments (given the status of this repository with over 1000 open issues I really wasn't sure if people are paying attention- so yes thank you very much!)

As it is, I am unsure if this repository is being actively managed, but it would be of great help to the community if there is a pretrained R-CNN model available on CLEVR. And I believe that based on your demo images, this could be the best one we have so far. 

I made some attempts to duplicate your listed procedures, but unfortunately, I got stuck on generating the CLEVR dataset with masks - possibly caused by some error related to Blender- also the generation of the images are time-consuming even with GPUs. Therefore, it could really be a time saver if a pretrained model is available.

Thank you again for the great work

rimamittal:
Here is a trained model (trained for 30 epochs):

https://drive.google.com/file/d/1NFcr9B3-ZbKv85Z4U_g0r4XZ0jP89TEY/view

Cuberick-Orion:
> Here is a trained model (trained for 30 epochs):
> 
> https://drive.google.com/file/d/1NFcr9B3-ZbKv85Z4U_g0r4XZ0jP89TEY/view

Thank you very much!

Cuberick-Orion:
> Here is a trained model (trained for 30 epochs):
> 
> https://drive.google.com/file/d/1NFcr9B3-ZbKv85Z4U_g0r4XZ0jP89TEY/view

Hi, I have encountered some errors in loading the model weights, specifically:
```python
ValueError: Layer #221 (named "mrcnn_bbox_fc"), weight <tf.Variable 'mrcnn_bbox_fc/kernel:0' shape=(1024, 388) dtype=float32> has shape (1024, 388), but the saved weight has shape (1024, 16).
```

In the `clevr.py` file there is the configuration of ` NUM_CLASSES = 1 + (3 * 2 * 8 * 2)
`, which corresponds to the current model configuration, as well as the number of classes in the CLEVR dataset. However, in the `inspect_clevr_model.ipynb` the `NUM_CLASSES` is 4 - in essence, it only accounts for the three shapes + background. I was not sure the reason for this inconsistency (as if, there are two versions of the model, one for detecting just the shape, the other for detecting all possible classes), could you help look into it? Thanks in advance

JensNevens:
I have encountered the same error when loading the model weights.
When loading the weights using
`model.load_weights(CLEVR_MODEL_PATH, by_name=True,exclude=["mrcnn_class_logits", "mrcnn_bbox_fc","mrcnn_bbox", "mrcnn_mask"])`
the model does not detect any objects.

Cuberick-Orion:
> I have encountered the same error when loading the model weights.
> When loading the weights using
> `model.load_weights(CLEVR_MODEL_PATH, by_name=True,exclude=["mrcnn_class_logits", "mrcnn_bbox_fc","mrcnn_bbox", "mrcnn_mask"])`
> the model does not detect any objects.

No, the correct way to avoid the error is to change the config file, specifically you need to change the `NUM_CLASSES` to 4, that way, the model is able to (at least) predict the shape of the objects.

rimamittal:
Hi Everyone,
This is correct, the model I shared is only trained for 4 classes. 

I did train the bigger model for all classes - shape, material, color etc  and unfortunately I don't have the model weights handy with me right now.
However, the code shared is sufficient enough to train the model on all the classes (change clevr.py accordingly depending on the number of classes).

I would be happy to re-train and share the bigger model, but would need a few weeks to respond.

Thank you

rimamittal:
Update: 

Please find the model trained on all classes(30 epochs) at the link below:

https://drive.google.com/file/d/16QsIzSCiT9zw0vkEinEfwePoR28d51jS/view?usp=sharing

Zhongping-Zhang:
Hi rimamittal,
The link isn't working now, would it be possible for you to post a new link for the pretrained model on CLEVR? Thank you very much!

JamesBatten:
The link to the trained model appears to have expired.

Is there an updated link?

Cuberick-Orion:
> 
> 
> The link to the trained model appears to have expired.
> 
> Is there an updated link?

Hi James, 

It's been a long time since I was working on a project related to this, but fortunately, the files are still on my drive.

Please see the below link to download [this checkpoint](https://github.com/matterport/Mask_RCNN/pull/1608#issuecomment-586072642) (the model trained on all classes(30 epochs)) :

https://drive.google.com/file/d/1MWdHLXnpqo7V8M35z1__11ETn97rSrE0/view?usp=sharing

JamesBatten:
Great! Thank you 😄

PR #1611, opened 2019-07-03
replaced data_generator function with a subclass of keras.utils.Sequence to avoid duplicate data when multiprocessing=True

world4jason:
#740 this one is too old for this version, and also it had some bugs.
I wrote a new one based on following tutorial.
https://stanford.edu/~shervine/blog/keras-how-to-generate-data-on-the-fly

btw, some of the solutions indicate that adding this line 
`
**multiprocessing.freeze_support()**   
`
before following line, will solve runtime error in Windows
`
workers = multiprocessing.cpu_count()
`


Ubikas:
@world4jason thanks for the data generator update, I was already thinking to do it myself. Anyway it has a small weakness. Probably it was designed and works when batch size is always equal: like 1 image in 1 batch. In my case I was running an experiment with 100 images. 4 GPU with 4 images each, 16 images per batch in total. The last batch gets only 4 images and generator runs into trouble. In the data_generator function I have changed:
`while b < self.batch_size:`
to
`while b < self.batch_size and b < image_ids.shape[0]:`

I guess it should help.

aaronzguan:
I think you forget to shuffle the image index at the start of an epoch.

world4jason:
@Ubikas my fault lol, thanks a lot

aaronzguan:
@world4jason Thanks for your update. I have tried your codes and it works. However, the training somehow stucks at 999 steps/1000 at epoch 100 just before the validation starts. I checked the CPU status and it seems the process is in the interruptible sleep status waiting for the IO. I used multiworkers, multiprocess with 8 GPUs. Do you have any idea how to fix it?

pnuehrenberg:
@world4jason thanks for the code!

I think when there is an image without instances
```python
if not np.any(gt_class_ids > 0):
    continue
```
the generator hangs at the current `b` because `b += 1` is never reached. Adding `b += 1` before `continue` should fix that.

Using this generator class smoothed my training loss curves dramatically, is that expected when the old generator function duplicated data when multiprocessing?

ustcjerry:
> @world4jason thanks for the code!
> 
> I think when there is an image without instances
> 
> ```python
> if not np.any(gt_class_ids > 0):
>     continue
> ```
> 
> the generator hangs at the current `b` because `b += 1` is never reached. Adding `b += 1` before `continue` should fix that.
> 
> Using this generator class smoothed my training loss curves dramatically, is that expected when the old generator function duplicated data when multiprocessing?

Not working for me, if i use this code below:
```
if not np.any(gt_class_ids > 0):
    b+=1
    continue
```
Just get errors like this:
`ValueError: output of generator should be a tuple (x, y, sample_weight) or (x, y). Found: None`

But the original code works fine in my case and the generator won't hang

valbertoenoc:
Guys, I don't know what else to do. First of all, thank you very much for your effort in solving this issue. Unfortunately, nothing worked for me. My training hangs at a random step during the first epoch. 

My setup is an i3 4130t with a Geforce 1050 Ti 4gb. 
tensorflow-gpu: 1.15.0
keras: 2.2.5

Someone got any light on that?

Thanks a lot.

EDIT: Works fine on CPU btw. But it's prohibitedly long.

ustcjerry:
@valbertoenoc  
maybe you can try to use a small number of images like 1 or 2 or 10 to test.
Your GPU only has 4gb which is low, which may cause some problems.

valbertoenoc:
@ustcjerry Thanks for the reponse.
I did put the minimum available. I also tried resizing the images and all. Nothing worked. 
I just have no idea what these freezing is. The system doesn't freeze at all, just the training process. It drives me crazy not knowing what is causing this at least.

travishsu:
Same issue, I solved it with this pull of keras https://github.com/keras-team/keras/pull/13322

x-jeff:
@world4jason  @Ubikas 
I've combined your two solutions.And it works.Thank you very much.

Adblu:
> while b < self.batch_siz

@Ubikas I would appreciate if you point precise location, because I cannot find it anywhere.

travishsu:
> > while b < self.batch_siz
> 
> @Ubikas I would appreciate if you point precise location, because I cannot find it anywhere.

https://github.com/world4jason/Mask_RCNN/blob/data_gen/mrcnn/model.py#L1707

Adblu:
@travishsu  Thank you. However this file significantly different than from https://github.com/matterport/Mask_RCNN/blob/master/mrcnn/model.py 
I guess that this version supports multiprocessing ?

If yes, shall i download all files from https://github.com/world4jason/Mask_RCNN/blob/data_gen/mrcnn/
in order to make it work ?

tanishqkancharla:
@valbertoenoc I know it's been a long time since you had that issue, but I was having the same problem. Turns out my issue was that I never actually installed this pull request! You have to uninstall the previous mask rcnn library `pip uninstall mask-rcnn` before installing this pull request.

bertsky:
Thanks @world4jason – very useful! I extended your solution exposing the generator for evaluation as well – [here](https://github.com/bertsky/Mask_RCNN/commit/9e1d5079b4d422d6d0e02fb51a572c2300c7090c).

Jeffin21:
Hey, tried adding the code of @world4jason @Ubikas got stuck in epoch 75 [](url) 
thalap
anyone please suggest how to deal with this ?

bertsky:
@Jeffin21 that's not epoch 75, but step/batch 75. I suspect your generator or file system has a problem. Try calling the generator function yourself 75 times and see what happens.

Jeffin21:
@bertsky am trying the balloon sample,

bertsky:
> Thanks @world4jason – very useful! I extended your solution exposing the generator for evaluation as well – [here](https://github.com/bertsky/Mask_RCNN/commit/9e1d5079b4d422d6d0e02fb51a572c2300c7090c).

See also [here](https://github.com/bertsky/Mask_RCNN/commit/e3115174ed5df1b38489c4412bd8b7880f3f2f3d) for an inference-time generator interface to facilitate CPU-side parallelism for faster detection.

However,

> The last batch gets only 4 images and generator runs into trouble. In the data_generator function I have changed:
> `while b < self.batch_size:`
> to
> `while b < self.batch_size and b < image_ids.shape[0]:`

that's not enough to get arbitrary batch sizes working: If `image_ids` is exhausted before `batch_size` is reached (i.e. in a partial/tail-batch), that generator will return `None` instead of the last batch, which will stall the queue eventually. You need to place the `return` statement outside of the loop and generalize its condition, too (as was done [here](https://github.com/bertsky/Mask_RCNN/commit/a02c408e02dccc7d25eae3cd12c12332997f2423)).

Jeffin21:
@bertsky thanks for your effort , would be great if you can provide a separate link of changed files, little confused in some areas on changing codes. thanks

PR #1642, opened 2019-07-17
Visualize.py Changed

Djrocks2105:
I am currently working on a project where i need to find the length-width of a vehicle and i am using the masks produced by Mask-RCNN .
So I have created a function in **visualize.py** file to get the mask of the required classes along with some other info which may come in handy for some other projects too.


Djrocks2105:
I am currently working on a project where i need to find the length-width of a vehicle and i am using the masks produced by Mask-RCNN .
So I have created a function in visualize.py file to get the mask of the required classes along with some other info which may come in handy for some other projects too.

PR #1649, opened 2019-07-21
Local api, can run generate_detections interactively with float32 imagery, testing with byte stream

rbavery:


rbavery:
accidental :)

PR #1807, opened 2019-10-18
Add upper bound for Tensorflow in requirements.txt

### LINKED ISSUE #1797: AttributeError: module 'tensorflow' has no attribute 'log'
i downloaded mask_rcnn_coco.h5 weights and installed pycocotools also and when i try to run demo.ipynb.  i am getting this error as AttributeError: module 'tensorflow' has no attribute 'log'
, please help i am still learning. I installed all the requoremnts.txt file also. please help.

---------------------------------------------------------------------------
**AttributeError                            Traceback (most recent call last)
<ipython-input-4-138183a2a830> in <module>
      1 # Create model object in inference mode.
----> 2 model = modellib.MaskRCNN(mode="inference", model_dir=MODEL_DIR, config=config)
      3 
      4 # Load weights trained on MS-COCO
      5 model.load_weights(COCO_MODEL_PATH, by_name=True)**

this is the error im getting

joshmcgrath08:
- Tensorflow 2.0 is not compatible with demo.ipynb due to removal of log function, potentially among others
- Fixes https://github.com/matterport/Mask_RCNN/issues/1797

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ Pillow
 cython
 matplotlib
 scikit-image
-tensorflow>=1.3.0
+tensorflow>=1.3.0, <2.0
 keras>=2.0.8
 opencv-python
 h5py


eamonnkenny:
actually I think that setting an upperbound is not the correct solution. I've just found the same issues.
tf.log should be correctly named tf.math.log
and tf.random_shuffle should be correctly named tf.random.shuffle for example.
The code is not being maintained any more or not being kept up to date as shown by:
https://github.com/tensorflow/tensorflow/issues/24653

Al-Badri179:
> actually I think that setting an upperbound is not the correct solution. I've just found the same issues.
> tf.log should be correctly named tf.math.log
> and tf.random_shuffle should be correctly named tf.random.shuffle for example.
> The code is not being maintained any more or not being kept up to date as shown by:
> [tensorflow/tensorflow#24653](https://github.com/tensorflow/tensorflow/issues/24653)

Where this change should be done, in which file?

PR #1847, opened 2019-11-04
Transformed model.py to tensorflow 2

alainkaiser:
Added a tensorflow-2 folder to the mrcnn folder.
The folder tensorflow-2 contains the transformed model.py file which can be used with tensorflow 2.


SalahAdDin:
What's about this?

PR #1886, opened 2019-11-28
Fix Keras engine topology to saving

marcelolandivar:
Fix Keras compatibility


SalahAdDin:
@marcelolandivar Why?

PR #2115, opened 2020-04-15
Compatible with Tensorflow 2.1

jiapei100:
Tested, works fine with TF 2.1


pbischoff:
TF 2.1 doesn't have tf.random_shuffle. Instead it uses tf.random.shuffle

ablaw123:
Thanks... This works well for inference but there are still errors when trying to train on a new dataset.

this is the error I get when trying to train.  This is after I made the change from  tf.random_shuffle. to  tf.random.shuffle


> tensorflow.python.framework.errors_impl.OperatorNotAllowedInGraphError: using a `tf.Tensor` as a Python `bool` is not allowed in Graph execution. Use Eager execution or decorate this function with @tf.function.

cl2227619761:
> Thanks... This works well for inference but there are still errors when trying to train on a new dataset.
> 
> 
> 
> this is the error I get when trying to train.  This is after I made the change from  tf.random_shuffle. to  tf.random.shuffle
> 
> 
> 
> 
> 
> > tensorflow.python.framework.errors_impl.OperatorNotAllowedInGraphError: using a `tf.Tensor` as a Python `bool` is not allowed in Graph execution. Use Eager execution or decorate this function with @tf.function.

Add tf.compat.v1.disable_eager_execution() in the first line will be ok

KossBoii:
> > Thanks... This works well for inference but there are still errors when trying to train on a new dataset.
> > this is the error I get when trying to train.  This is after I made the change from  tf.random_shuffle. to  tf.random.shuffle
> > > tensorflow.python.framework.errors_impl.OperatorNotAllowedInGraphError: using a `tf.Tensor` as a Python `bool` is not allowed in Graph execution. Use Eager execution or decorate this function with @tf.function.
> 
> Add tf.compat.v1.disable_eager_execution() in the first line will be ok

@cl2227619761 which file and where do you put this line of code?

PR #1896, opened 2019-12-08
Tensorflow 2.0 compatibility

tomgross:
Things to do / fix
- [x] training mode
- [x] inference mode
- [x] calling crop_and_resize_v1 (from tensorflow.python.ops.image_ops_impl) with box_ind is deprecated and will be removed in a future version.
Instructions for updating:
box_ind is deprecated, use box_indices instead
- [x] metrics
- [x] all provided samples run (demo, training_shape, balloon)
- [x] add some tests
- [x] maybe split PR for better reviewability

To try this do the following:

## Install Mask_RCNN
```
git clone https://github.com/tomgross/Mask_RCNN.git --branch tensorflow-2.0
cd Mask_RCNN
bin/pip install --upgrade pip
bin/pip install -r requirements.txt
bin/pip install -e .
```

## Run examples

There are some examples for training and inference mode you can run as notebooks in your setup.

## Run tests

```
bin/pip install pytest
bin/pytest tests
```

## Known issues

- Tensorflow Version 2.1.0 seems to have troubles as reported here: https://github.com/matterport/Mask_RCNN/pull/1896#issuecomment-570299903 I did not run into troubles with the version 2.0.0 so far. If you encounter issues with 2.1.x please report, help fix or use 2.0.0
- numpy version 1.18.0 seems to conflict with imgaug. Downgrading to 1.17.4 works for me.
- If you use  `load_image_gt` in your code/notebook remove the `use_mini_mask`argument since it has been use redundant with config: load_image_gt(dataset, config, image_id<del>, use_mini_mask=False<del>)
- Older versions of pip don't find the releases of tensorflow 2. I don't know which is the breaking version, but `19.3.1` works.

PR #2209, opened 2020-05-27
running on tensorflow 2.2 on Windows10

leekunhee:
code based on #1896 
I made a constant layer for **"A hack to get around Keras's bad support for constants"**
it works well on Windows 10, Python 3.7, Tensorflow 2.2

the code below is not working on tf.keras. It tells me to create and use a new layer class.
```
anchors = tf.keras.layers.Lambda(lambda x: tf.Variable(anchors), name="anchors")(input_image)
```

```
class ConstLayer(tf.keras.layers.Layer):
    def __init__(self, x, name=None):
            super(ConstLayer, self).__init__(name=name)
                self.x = tf.Variable(x)

            def call(self, input):
                return self.x

anchors = ConstLayer(anchors, name="anchors")(input_image)
```


akTwelve:
I'm not a maintainer of the repo, but I've tested this branch and it seems to work great with Windows 10, CUDA 10.1, and Tensorflow 2.2. Thank you @leekunhee!

mrcnn/config.py
```diff
@@ -227,10 +227,17 @@ def __init__(self):
...
+
     def display(self):
         """Display Configuration values."""
         print("\nConfigurations:")
-        for a in dir(self):
-            if not a.startswith("__") and not callable(getattr(self, a)):
-                print("{:30} {}".format(a, getattr(self, a)))
+        for key, val in self.to_dict().items():
+            print(f"{key:30} {val}")
+        # for a in dir(self):
```
akTwelve: I'd suggest removing this commented out code, or explaining why you left these lines in

mrcnn/model.py
```diff
@@ -8,33 +8,34 @@
...
 import tensorflow as tf
-import keras
-import keras.backend as K
-import keras.layers as KL
-import keras.engine as KE
-import keras.models as KM
+import tensorflow.keras as keras
+import tensorflow.keras.backend as K
+import tensorflow.keras.layers as KL
+import tensorflow.keras.layers as KE
```
akTwelve: This line should probably be removed, since there are only a few places that reference KE and and they can all be updated to KL

mrcnn/model.py
```diff
@@ -1931,7 +1929,22 @@ def build(self, mode, config):
             # TODO: can this be optimized to avoid duplicating the anchors?
             anchors = np.broadcast_to(anchors, (config.BATCH_SIZE,) + anchors.shape)
             # A hack to get around Keras's bad support for constants
-            anchors = KL.Lambda(lambda x: tf.Variable(anchors), name="anchors")(input_image)
+            # 원본 tf1 버전 코드
```
akTwelve: Perhaps you could include an English translation of your notes.

requirements.txt
```diff
@@ -4,8 +4,8 @@ Pillow
 cython
 matplotlib
 scikit-image
-tensorflow>=1.3.0
-keras>=2.0.8
+tensorflow==1.13.1
```
akTwelve: Isn't the whole point of this change to support tensorflow 2.0+? 🙂

requirements.txt
```diff
@@ -4,8 +4,8 @@ Pillow
...
+keras==2.0.8
```
akTwelve: I think keras is safe to remove, since all instances are part of the tensorflow module

leekunhee: It's my first try, so I make a lot of mistakes. Thank you so much for your review :)

kiflowb777:
Please consider replacing KL.Lambda layers with custom_layers in TF2.x
More info:
https://github.com/matterport/Mask_RCNN/issues/2494

PR #2309, opened 2020-08-07
Update requirements.txt

muhendis:
Can you update the requirements.txt file for stable versions of libraries for the mask r-cnn ? (example error: # 1754)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,8 +4,8 @@ Pillow
 cython
 matplotlib
 scikit-image
-tensorflow>=1.3.0
-keras>=2.0.8
+tensorflow==1.6.0
+keras== 2.2.0
 opencv-python
 h5py
 imgaug


shining-love:
> Can you update the requirements.txt file for stable versions of libraries for the mask r-cnn ? (example error: # 1754)

The requirements.txt file is clearing,  It's not  necessary to fix the  versions of libraries for the mask r-cnn.

muhendis:
i think it is necessary because the problem takes time. And many people faced this problem.Am i wrong ?

shining-love:
> i think it is necessary because the problem takes time. And many people faced this problem.Am i wrong ?

No, It is so kind of you.Thanks for your works!

muhendis:
thanks, good luck! @yang-gis

Shakesbeer333:
pycocotools might be missed

PR #2317, opened 2020-08-13
Add save detection result as image function to visualize.py

ap3885:


ap3885:
Base code taken from https://github.com/matterport/Mask_RCNN/blob/bc8f148b820ebd45246ed358a120c99b09798d71/visualize.py and modified some errors I was getting.

Issue #817, opened 2018-07-27
InvalidArgumentError: slice index 33 of dimension 0 out of bounds at the time of prediction


janismdhanbad:
I trained my model successfully and now I am trying to predict on images. I am specifying "IMAGES_PER_GPU" in InferenceConfig class. Till 32 "IMAGES_PER_GPU" there is no error but after that, it's giving me the error below. I checked some issues on tensorflow repository and I think it might be related to the input that is being passed. Also, the below error is giving an error in line 828 of utils.py

**inputs_slice = [x[i] for x in inputs]** 

InvalidArgumentError                      Traceback (most recent call last)
~/anaconda3/envs/tensorflow_p36/lib/python3.6/site-packages/tensorflow/python/client/session.py in _do_call(self, fn, *args)
   1321     try:
-> 1322       return fn(*args)
   1323     except errors.OpError as e:

~/anaconda3/envs/tensorflow_p36/lib/python3.6/site-packages/tensorflow/python/client/session.py in _run_fn(feed_dict, fetch_list, target_list, options, run_metadata)
   1306       return self._call_tf_sessionrun(
-> 1307           options, feed_dict, fetch_list, target_list, run_metadata)
   1308 

~/anaconda3/envs/tensorflow_p36/lib/python3.6/site-packages/tensorflow/python/client/session.py in _call_tf_sessionrun(self, options, feed_dict, fetch_list, target_list, run_metadata)
   1408           self._session, options, feed_dict, fetch_list, target_list,
-> 1409           run_metadata)
   1410     else:

InvalidArgumentError: slice index 33 of dimension 0 out of bounds.
	 [[Node: ROI_1/strided_slice_228 = StridedSlice[Index=DT_INT32, T=DT_FLOAT, begin_mask=0, ellipsis_mask=0, end_mask=0, new_axis_mask=0, shrink_axis_mask=1, _device="/job:localhost/replica:0/task:0/device:GPU:0"](_arg_input_anchors_1_0_0/_8681, mrcnn_detection_1/strided_slice_744/stack_1, mrcnn_detection_1/strided_slice_767/stack_1, mrcnn_mask_deconv_1/strided_slice/stack_1)]]
[...]

PR #2367, opened 2020-09-11
Fix slice index out of bounds for batch_size >32 #817

Dawars:
The default **batch_size** for `keras_model.predict()` is 32 therefore the code will crash with a "InvalidArgumentError: slice index 33 of dimension 0 out of bounds" error if batch_size is more than that.

This fix is based on the issue https://github.com/matterport/Mask_RCNN/issues/817

### DIFF
--- a/mrcnn/model.py
+++ b/mrcnn/model.py
@@ -2521,7 +2521,7 @@ class MaskRCNN():
             log("anchors", anchors)
         # Run object detection
         detections, _, _, mrcnn_mask, _, _, _ =\
-            self.keras_model.predict([molded_images, image_metas, anchors], verbose=0)
+            self.keras_model.predict([molded_images, image_metas, anchors], batch_size=len(images), verbose=0)
         # Process detections
         results = []
         for i, image in enumerate(images):


bertsky:
Well spotted! The `bach_size=32` fallback is still effective with current Keras. This means that the current master version would fail if `IMAGES_PER_GPU * GPU_COUNT > 32`, and (of course) that `detect` is unnecessarily clumsy to use on a bunch of samples – you have to split them up "by hand" instead of letting `keras_model.predict` do the work. 

To address the latter problem – which still persists with this PR – I propose attacking the problem from the other side: Why not instead pass `batch_size=config.BATCH_SIZE` to `keras_model.predict()`? With that we could get rid of …

https://github.com/matterport/Mask_RCNN/blob/3deaec5d902d16e1daf56b62d5971d428dc920bc/mrcnn/model.py#L2494-L2495

… and call with arbitrarily long (or short) lists of images. The only missing ingredient would be a change to the batch dimension of `anchors` (currently just broadcast to batch_size – would then have to be broadcast to the number of images IIUC).

PR #2368, opened 2020-09-18
requirements file should update to exact version.

Oscarjia:
In fact, this project cannot run normally now. For example, there is no version of tensorflow 1.3.So I updated the requirements to make this project run.


laygond:
@Oscarjia Your requirement file as it is, gave me several installation problems. However, I took your versions as guidance and everything worked. Thank you.  
```
numpy==1.18.5
scipy==1.1.0
Pillow==7.2.0
cython==0.29.21
matplotlib==3.3.2
scikit-image==0.17.2
tensorflow-gpu==1.15.0
keras==2.0.8
opencv-python==4.4.0.42
h5py==2.10.0
imgaug==0.4.0
IPython[all]
```

CLKVoss:
> 
> 
> In fact, this project cannot run normally now. For example, there is no version of tensorflow 1.3.So I updated the requirements to make this project run.

Works well with python 3.7

PR #2615, opened 2021-06-27
KE.Layer ==> KE.base_layer.Layer

jordan-bonecutter:
KE.Layer ==> KE.base_layer.Layer to reflect keras api change


3F3Feng (approved):
Thanks for fixing this

jiubie12:
Do as you said, but it appears：AttributeError: module 'tensorflow' has no attribute 'log'

PR #2855, opened 2022-07-13
Bug/fix parallel mode

svbeuningen:


svbeuningen:
1. made number of workers and multiprocessing variable in train
2. Fixed parallel mode for keras 2.2.5

PR #2888, opened 2022-10-01
Layer error fix

ankushKun:
Fixes issue #2783 
image

Due to updates in Keras library we change `KE.Layer` with `KL.Layer` in multiple locations in `mrcnn/model.py` file


MrWalu (approved):
One letter makes all the differences. Good job

